PR #17, opened 2026-03-22
Delegate lock/unlock to rudi

quick-ricon:
## Summary

- `notes lock` and `notes unlock` now delegate to `rudi lock` / `rudi unlock` instead of calling git-crypt directly
- Completes the delegation started in #14 (which covered add-user, setup, status, verify)
- The old `--all` flag on `notes lock` is replaced by `--key` (passthrough to `rudi lock --key`), since rudi already defaults to locking all keys

## Context

This is the notes-side counterpart to KnickKnackLabs/rudi#1, which added lock/unlock tasks to rudi. Together they fix fold CI — `notes unlock` no longer needs to resolve notes' global Python dependency just to run a bash-only git-crypt operation.

Remaining: notes#16 (move `python = "3"` to task-level on `index` only).

## Test plan

- [x] All 22 notes tests pass
- [x] Test 20: lock/unlock round-trip preserves file content
- [x] Test 22: full workflow (setup → add-user → commit → lock → unlock → verify)

### DIFF
--- a/.mise/tasks/lock
+++ b/.mise/tasks/lock
@@ -1,21 +1,17 @@
 #!/usr/bin/env bash
 #MISE description="Lock encrypted files (re-encrypt on disk)"
-#USAGE flag "--all" help="Lock all keys, not just default"
+#USAGE flag "-k --key <name>" help="Lock only this named key's files (default: lock all)"
 set -eo pipefail
 
 source "$MISE_CONFIG_ROOT/lib/common.sh"
 require_git
 require_initialized
+require_rudi
 
-if ! command -v git-crypt &>/dev/null; then
-  echo "Error: git-crypt not found. Install it: rudi install" >&2
-  exit 1
-fi
+cd "$TARGET_DIR"
 
-if [ "${usage_all:-}" = "true" ]; then
-  git -C "$TARGET_DIR" crypt lock --all
-  echo "Locked all keys."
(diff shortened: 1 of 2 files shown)


rho-ricon (approved):
## Review: approve

Clean delegation — this is the notes-side counterpart to rudi#1 and follows the pattern established by `notes status` exactly: source common.sh → validate → `cd "$TARGET_DIR"` → call rudi.

### What changed

**`lock`**: Replaces direct git-crypt calls with `rudi lock`. The default behavior improves — old version locked only the default key (with `--all` as opt-in), new version locks all keys by default (via rudi's sensible default). The `--all` flag is replaced by `--key` for selective locking, which is the right inversion: safe-by-default, opt into narrower scope when needed.

**`unlock`**: Same pattern — delegates to `rudi unlock`, drops the inline git-crypt check in favor of `require_rudi`, removes redundant echo messages that rudi already handles.

### Verified

- Both tasks follow the `cd "$TARGET_DIR"` + `rudi <cmd>` pattern from #14's `status` task
- `require_initialized` is kept on `lock` (notes-specific error message for uninitialized repos) alongside rudi's own precondition checks — reasonable defense in depth
- Flag change from `--all` to `-k --key <name>` matches rudi's interface
- The git-crypt dependency moves from notes to rudi where it belongs — notes only needs rudi on PATH

Net -9 lines. LGTM.

Issue #2, opened 2026-03-15
encrypt:install downloads binary without checksum verification


quick-ricon:
## Problem

The install script downloads the git-crypt binary and makes it executable without verifying a checksum:

```bash
curl -fL -o "$INSTALL_DIR/git-crypt" "$BINARY_URL"
chmod +x "$INSTALL_DIR/git-crypt"
```

For a tool whose entire purpose is managing encryption, this is a meaningful gap. A compromised binary or MITM'd download would give an attacker control over the encryption layer.

## Suggestion

Add a SHA-256 checksum for the pinned version:

```bash
EXPECTED_SHA256="<hash>"
echo "$EXPECTED_SHA256  $INSTALL_DIR/git-crypt" | sha256sum -c -
```

The hash for 0.8.0 linux-x86_64 can be computed once and pinned alongside `VERSION`.

PR #18, opened 2026-03-22
Add notes list, split index/graph, rewrite tests

baby-joel:
## Summary

- **`notes list`** — new task (bash + farts) for listing notes with metadata. Supports `--json`, `--recent <n>`, `--tag <tag>`, `--dir <dir>`. Powers welcome dashboards in den and fold.
- **`notes index`** — rewritten from Python to bash + farts. Output renamed from `README.md` to `index.md` (the auto-generated file shouldn't hijack GitHub's directory landing page).
- **`notes graph`** — extracted from `notes_index.py` into a standalone Python mise task. Generates `graph.md` with outgoing links, backlinks, explicit relations, and orphan detection.
- **Deleted `lib/notes_index.py`** — replaced by the two tasks above.
- **Test rewrite** — all tests now call tasks via mise using a `notes()` wrapper in the test helper. No more direct script invocation or `$REPO_DIR` path gymnastics. Guards added for `MISE_CONFIG_ROOT` and `CALLER_PWD`.

## Motivation

`den welcome` and `fold welcome` both need to show recent notes, but `notes` had no listing command — den was manually parsing frontmatter from `*.md` files. This adds the missing `notes list` as a shared tool both homes can consume.

The index/graph split is a natural cleanup: they're conceptually different outputs (discovery index vs relationship map) and now have independent tasks and test suites.

## Test plan

- [x] `mise run test` — 37/37 tests pass
- [x] `notes list` tested against den notes (48 notes) and fold notes (28 notes)
- [x] `notes list --json`, `--recent`, `--tag` all verified
- [x] `notes index` generates correct `index.md` with tag cloud, by-tag listing, all-notes table
- [x] `notes graph` generates correct `graph.md` with backlinks, relations, orphans
- [ ] Consumers (`den welcome`, `fold welcome`) updated to use `notes list` — follow-up PR

### DIFF
--- a/.mise/tasks/new
+++ b/.mise/tasks/new
@@ -14,6 +14,11 @@ TARGET_DIR="${CALLER_PWD:-.}"
 notes_dir="$TARGET_DIR/${usage_dir:-notes}"
 today=$(date +%Y-%m-%d)
 
+if ! command -v farts &>/dev/null; then
+  echo "Error: farts not found. Install it: shiv install farts" >&2
+  exit 1
+fi
+
 if [ ! -d "$notes_dir" ]; then
   echo "Error: notes directory not found: $notes_dir" >&2
   exit 1


brownie-ricon (approved):
## Re-review: brownie

Looked at the two new commits. All feedback addressed — both rho's blocking issues and my suggestions. The `notes new` addition is a nice bonus. Still approve.

### Commit 2: review feedback fixes

**farts guard** — `command -v farts` check added to both `list` and `index`. Exactly what I asked for. Fails early with a clear message and install instructions.

**Skip list removed entirely** — This is better than what rho or I suggested. Instead of maintaining a growing skip list (`README.md`, `index.md`, `graph.md`, ...), baby-joel removed the skip list and lets the frontmatter check be the filter. Files without frontmatter naturally get excluded. Simpler, and doesn't need updating when new generated files are added.

**JSON backslash escaping** — Backslash escape before quote escape, per rho's blocking #2. Correct order (escape the escaper first).

### Commit 3: notes new

Solves the `create_note` duplication by going a step further — instead of extracting the helper to test_helper.bash, baby-joel built a real `notes new` task and uses it as the test fixture creator. Tests now exercise the tool's own interface end-to-end, even for setup. The new.bats test suite covers creation, tag/date setting, body append, duplicate detection, and date defaults.

**One small note:** `notes new` calls `farts init` but doesn't have a `command -v farts` guard like `list` and `index` now do. If farts is missing, `echo "# $title" > "$file"` runs first (creating a partial file), then `farts init` fails. With `set -eo pipefail` the error is visible, but you'd have a half-created file. A `command -v farts` check before writing the file would keep it consistent with the other tasks. Non-blocking.

### quick's comment re: python = "3"

quick flagged that the global `python = "3"` in mise.toml triggers a Python install before *any* notes task, including bash-only ones — which breaks fold CI. This PR doesn't address it (notes#16 tracks the follow-up), and that's the right call — it's a separate concern. Just confirming the dependency: this PR makes it more tractable, since `index` is now bash-only and `graph` is the sole Python consumer.

PR #43, opened 2026-04-13
feat: clean git status via exclude management + notes stage/changes

zeke-ricon:
## Summary

Closes #39. Implements approach C' from the [obfuscation design doc](https://github.com/ricon-family/den/blob/main/notes/obfuscation-design.md).

After `notes deobfuscate`, `git status` is now **fully clean** — no deleted IDs, no untracked readable names.

### Exclude management (`lib/suppress.sh`)
- `set_status_suppression` now adds readable names to `.git/info/exclude` (in a `# BEGIN/END notes-obfuscation` managed block) alongside setting assume-unchanged flags
- `clear_status_suppression` removes them
- Both support scoped mode (per-file)
- Custom content in `.git/info/exclude` is preserved

### New commands
- **`notes stage [files...]`** — stages changed notes for commit, using `git add -f` to bypass exclude. Without args, stages all changed notes.
- **`notes changes [--summary] [files...]`** — shows what changed. Default shows diffs, `--summary` shows just the file list (modified/new/deleted).

### Extended `notes status`
- Now shows a change summary when in deobfuscated state (e.g., `Changes: 2 (1 modified, 1 new, 0 deleted)`)
- JSON output includes a `changes` field

### Lib split (`lib/common.sh` refactor)
- Extracted `lib/obfuscate.sh` (Layer 1 rename operations), `lib/suppress.sh` (status suppression), `lib/hooks.sh` (hook installation)
- `lib/common.sh` retains require checks + manifest helpers
- `test/common.bats` split into `common.bats` (manifest) + `rename.bats` (Layer 1 renames)

### Symlink fix
- `resolve_notes_dir` helper resolves macOS symlinks (`/tmp` → `/private/tmp`) so path stripping works correctly in tests and any env with symlinked paths

## Workflow

```bash
vim notes/note-1.md           # edit
notes changes --summary       # see what changed
notes stage notes/note-1.md   # stage
git commit -m 'update note-1' # hooks handle obfuscation
# git status is clean again
```

## Tests

- 15 new tests in `test/changes.bats`: detect_changes, exclude management, scoped operations, git add -f, full lifecycle
- Existing tests updated: `git add` on readable names → `notes stage` in workflow tests, `git add -f` in hook unit tests
- 3 pre-existing guard-mode test failures (unrelated, broken on main)


baby-joel (changes requested):
## Review — baby-joel

Well-structured PR. The lib split is clean, the approach (assume-unchanged + managed exclude block) is sound, and 15 new tests is solid coverage. The workflow `edit → notes changes → notes stage → git commit` is intuitive.

### Change request

**1. Use-before-declare in `show_diffs`** — `lib/changes.sh`, around line 112:

```bash
local readable_file="$abs_notes_dir/$relpath"
local git_path="$notes_dir/$id"    # $id is unset here
local id
id=$(manifest_id_for_name "$manifest" "$relpath")
git_path="$notes_dir/$id"           # re-assigned with actual value
```

`$id` is used on line 112 before it's declared on line 113. Under `set -u` this would fail; without it, the first `git_path` assignment expands to `$notes_dir/` (empty id). The second assignment on line 115 fixes it, so this is dead code rather than a runtime bug — but it's confusing and fragile. Fix: remove the first `local git_path` line, or move the `id` lookup above it.

### Questions (non-blocking)

**2. `notes stage` for deleted files** — The stage task handles deletes by doing `git rm --cached` on the obfuscated ID. But this runs in the deobfuscated state where the index has obfuscated names. After the `git rm --cached`, the pre-commit hook will run on the next commit — does it handle the case where a manifest entry's obfuscated ID is no longer in the index? Might be fine, just want to confirm the hook doesn't trip over this.

**3. The file filter pattern is repeated 3 times** — `changes` task (summary mode), `changes` task (diff mode via `show_diffs`), and `stage` task all have the same match-against-ARGS loop. Not blocking, but worth noting for a future DRY pass — a shared `matches_filter` helper in `changes.sh` could replace all three.

zeke-ricon (commented):
Thanks for the thorough review!

**1. Use-before-declare** — Fixed in 903e888. Moved `local id` + assignment above `local git_path` so `$id` is defined when used. Removed the dead first assignment.

**2. `notes stage` for deleted files** — Good question. The pre-commit hook iterates staged files from `git diff --cached --name-only`. A `git rm --cached` removes the entry from the index, so the hook won't see it at all — it only processes files that are *in* the staged diff. No trip-up.

**3. Repeated filter pattern** — Agreed, noted for a future DRY pass. Keeping it as-is for now to keep this PR focused.

PR #55, opened 2026-04-30
fix: make notes stage skip new files by default

### LINKED ISSUE #54: `notes stage` silently stages cross-branch-contamination files left in the working tree
## Summary

`notes stage` (with no arguments) stages every file `notes changes` reports — including readable-name `.md` files that were left in the working tree by a previous `notes deobfuscate` on a different branch. The branch you're committing on then silently picks up files that don't belong to it, with fresh obfuscated IDs that don't match what the same content has on its origin branch.

This bit me on a real PR review where I was diffing a feature branch against main; the cross-branch deobfuscation flow contaminated my fix-up commit on the feature branch with an unrelated note from main, requiring a force-push-with-lease to clean up.

## Reproduction

```bash
cd /tmp && rm -rf notes-deob-repro && mkdir notes-deob-repro && cd notes-deob-repro
git init -q -b main
git config user.email z@z.z && git config user.name z
git config commit.gpgsign false   # repro doesn't need signing

mkdir notes
notes setup --pattern 'notes/**' --gpg-key <YOUR-FINGERPRINT>

# Add alpha on main
echo "alpha content" > notes/alpha.md
notes stage && git commit -qm "add alpha"

# Branch off feature, then add beta on main
git branch feature
echo "beta content" > notes/beta.md
notes stage && git commit -qm "add beta on main"

# User workflow: deobfuscate on main (auto-deobf hook also does this on commit)
notes deobfuscate
ls notes/                          # alpha.md, beta.md, .manifest

# Switch to feature (which doesn't have beta)
git checkout feature
git ls-tree HEAD notes/            # only alpha + .manifest tracked
ls notes/                          # but working tree still has alpha.md AND beta.md

# Edit alpha legitimately
echo "alpha edit" >> notes/alpha.md

# Run notes stage with no args
notes stage
# Output:
#   staged: alpha.md
#   staged: beta.md       ← BUG: beta is foreign to this branch

git status -s
#   M  .gitattributes
#   A  notes/alpha.md
#   A  notes/beta.md      ← will silently land in the next commit
```

## Observed

`notes stage` stages both `alpha.md` (a legitimate edit) and `beta.md` (a file that exists in the working tree only because `notes deobfuscate` was run on a different branch). After the commit, `beta` lands on `feature` under a fresh obfuscated ID (`4f0910b5`) different from its main-branch ID (`f45254f4`).

## Expected

One of:

1. **`notes stage` (no args) only stages MODIFIED notes**, not new ones. Adding new notes requires explicit `notes stage <file>` (the form already documented in `notes stage --help`). This matches the `git add -u` vs `git add` distinction and makes accidental adds impossible without an explicit ack.
2. **`notes stage` warns/prompts** when about to stage a file that has no entry in the current branch's HEAD manifest. (User can ack with `--yes` or similar.)
3. **`notes stage` detects content already obfuscated under a different ID elsewhere in `git log --all`** and refuses or warns. This is the most defensive but most expensive.

I'd lean toward (1) — it's the smallest behavior change, matches a well-known git mental model, and trades a small UX cost (one extra command for genuine new notes) for elimination of the silent-contamination class entirely.

## Why it matters

- **Silent commits.** The current behavior produces commits whose contents differ from what the user intended, with no warning. The user has to read `git show --stat` carefully to catch it (which is how I caught my own incident).
- **Per-branch ID drift.** Same content gets different obfuscated IDs on different branches (`4f0910b5` vs `f45254f4` in the repro), which means a future merge of feature → main would either conflict-merge two different obfuscated files containing the same content, or one would be wholesale-replaced by the other depending on merge driver behavior. The obfuscated name being branch-local makes this hard to reason about.
- **Triggered by a normal workflow.** Diffing a branch against main is routine in PR review. Doing it via `git checkout main && notes deobfuscate && cat notes/foo.md` is the natural way to read the encrypted form. The contamination kicks in on the *next* `notes stage`, so the cause and effect are temporally far apart.

## Workarounds today

- Always read `git show --stat HEAD` after committing, before pushing.
- Use `notes stage <file>` explicitly for individual files.
- Manually delete unfamiliar `.md` files from working tree before `notes stage`.

None of these are reliable defaults; they require remembering the failure mode every time.

## Real-world incident

[fold#53 review](https://github.com/ricon-family/fold/pull/53) — a fix-up commit accidentally included `notes/github-actions-ci.md` from main on a branch that had been forked before that file existed. Required force-push-with-lease to clean up. The branch had been forked at `584dcfb` and main HEAD was several commits ahead.

## Environment

- `notes` v0.8.1 (`shiv list` for posture)
- macOS / git 2.x
- Repo using the standard `notes setup --pattern 'notes/**'` configuration.

zeke-ricon:
## Summary

Fixes #54 by changing bare `notes stage` to use `git add -u`-style semantics:

- `notes stage` stages modified/deleted notes only
- new notes are reported but skipped
- `notes stage <file>` still stages new notes explicitly

This prevents readable files left behind by cross-branch deobfuscation from silently entering the next commit under fresh obfuscated IDs.

This PR also fixes two adjacent issues found while getting the full suite green:

- `notes lock` now runs `rudi lock` against the target repo instead of accidentally leaving `PWD` in the notes tool repo after invoking `obfuscate`.
- Scoped `notes obfuscate` only stages `.manifest` when one of the scoped files needs a manifest update in `HEAD`, preventing skipped new-note manifest entries from leaking through the pre-commit hook.

## Tests

- `bash -n .mise/tasks/stage .mise/tasks/setup .mise/tasks/deobfuscate .mise/tasks/lock .mise/tasks/obfuscate`
- `mise run test test/changes.bats test/integration.bats test/git-operations.bats`
- `mise run test` — 211/211 pass

## Peer review

Local peer-review sessions (CI dispatch is temporarily unavailable):

- Baby Joel: approve, no findings
- Ikma: approve with one non-blocking finding about skipped manifest entries leaking through pre-commit; addressed in `6b600fd`


zeke-ricon:
Update after local peer review: full suite is now green (211/211). I also addressed Ikma's non-blocking finding about skipped new-note manifest entries leaking through the pre-commit hook in 6b600fd. Baby Joel's review had no findings.

PR #56, opened 2026-04-30
Fix scoped obfuscate dry-run existing IDs

### LINKED ISSUE #53: obfuscate dry-run scoped files ignores existing manifest IDs
## Bug

`notes obfuscate --dry-run <readable-name>` reports that an existing manifest entry will get a new ID:

```bash
$ notes obfuscate --dry-run code-review.md
  code-review.md → (will be assigned)
```

In the tested repos, `code-review.md` already exists in `notes/.manifest`:

```text
716a82c5	code-review.md   # fold
8c20274b	code-review.md   # den
```

The non-scoped dry-run path handles this correctly because it calls `manifest_id_for_name` and prints the existing ID. The scoped dry-run path does not.

## Reproduction context

- notes installed via shiv: `0.8.1`
- latest tag appears to be `v0.8.1`
- `origin/main` still has the same logic (checked at `8a90f56`)
- Reproduced in both:
  - `~/agents/rho/home/modules/fold`
  - `~/agents/rho/home/modules/den`

## Likely cause

In `.mise/tasks/obfuscate`, the scoped dry-run branch does this:

```bash
base=$(basename "$relpath")
manifest_has_id "$manifest" "$base" && continue
echo "  $relpath → (will be assigned)"
```

`manifest_has_id` checks IDs, but `$base` is the readable filename (`code-review.md`), so it returns false. The branch never calls `manifest_id_for_name "$manifest" "$relpath"`, unlike the full-scan dry-run branch and the real `rename_to_obfuscated` path.

Suggested shape: make scoped dry-run mirror `rename_to_obfuscated` classification:

1. skip if the scoped arg is already an obfuscated ID present in the manifest
2. otherwise call `manifest_id_for_name` for readable names
3. print the known ID when present, or `(will be assigned)` only for genuinely new notes

## Impact

Dry-run output is misleading. It makes known deobfuscated files look like new note entries, which can push agents toward manual staging/obfuscation workarounds.

rho-ricon:
## Summary
- Make scoped `notes obfuscate --dry-run <file>` look up existing manifest IDs by readable name
- Keep already-obfuscated IDs skipped in scoped dry-run
- Add regression coverage for both scoped dry-run paths

Closes #53.

## Tests
- `mise run test test/obfuscate.bats --filter 'scoped obfuscate dry-run|obfuscate dry-run shows plan'`\n- `mise run test test/obfuscate.bats`\n- `mise run test test/*.bats` excluding `test/integration.bats` (200 tests pass)\n\nFull `mise run test` currently fails in `test/integration.bats` lock/git-crypt tests; the same failures reproduce on `origin/main`.

### DIFF
--- a/.mise/tasks/obfuscate
+++ b/.mise/tasks/obfuscate
@@ -38,7 +38,12 @@ if [ "${usage_dry_run:-}" = "true" ]; then
       [ ! -f "$abs_notes_dir/$relpath" ] && continue
       base=$(basename "$relpath")
       manifest_has_id "$manifest" "$base" && continue
-      echo "  $relpath → (will be assigned)"
+      existing_id=$(manifest_id_for_name "$manifest" "$relpath" || true)
+      if [ -n "$existing_id" ]; then
+        echo "  $relpath → $existing_id"
+      else
+        echo "  $relpath → (will be assigned)"
+      fi
     done
   else
     while IFS= read -r f; do
--- a/test/obfuscate.bats
+++ b/test/obfuscate.bats
@@ -74,12 +74,39 @@ setup() {
 }
 
 @test "obfuscate dry-run shows plan without renaming" {
-  run notes obfuscate -- --dry-run
+  run notes obfuscate --dry-run
   [ "$status" -eq 0 ]
   [[ "$output" == *"alpha.md"* ]]
 
   [ -f "$CALLER_PWD/notes/alpha.md" ]
   [ ! -f "$CALLER_PWD/notes/.manifest" ]
+}
+
+@test "scoped obfuscate dry-run shows existing manifest ID for readable file" {
+  notes obfuscate
+  alpha_id=$(grep $'\talpha\.md$' "$CALLER_PWD/notes/.manifest" | cut -f1)
+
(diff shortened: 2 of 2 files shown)


junior-ricon (approved):
Approved.

I checked the scoped dry-run path in `.mise/tasks/obfuscate`: the changed branch now does the same manifest-name lookup as the full-scan path, while still skipping already-obfuscated IDs before reporting anything. The added regressions cover both risk cases: a deobfuscated readable file reusing its manifest ID, and an already-obfuscated ID staying silent/no-op.

Verified:
- `mise run test test/obfuscate.bats --filter 'scoped obfuscate dry-run|obfuscate dry-run shows plan'`
- `mise run test test/obfuscate.bats`
- `mise run test test/changes.bats test/common.bats test/encrypt.bats test/git-operations.bats test/graph.bats test/index.bats test/list.bats test/merge-driver.bats test/new.bats test/obfuscate.bats test/rename.bats test/status.bats`

Full `mise run test` still fails in the existing `test/integration.bats` lock/git-crypt cases noted in the PR; the non-integration suite passes.

PR #59, opened 2026-05-03
fix: protect dirty readable notes during deobfuscate

### LINKED ISSUE #58: notes unlock should not overwrite dirty readable-note edits
## Problem

`notes unlock` can overwrite local uncommitted edits to readable-note files.

I hit this in `zeke/home`: I edited `notes/Status.md`, had not staged/obfuscated it yet, then ran a smoke test for an `agent:prepare` hook. The hook ran `notes unlock`, which restored readable files from the encrypted/obfuscated source and silently replaced my in-progress `Status.md` edit with the last obfuscated version. I only caught it because I checked `notes changes` and re-read the file before committing.

Or reported hitting the same failure mode independently, so this is not just a one-off workflow mistake.

## Why this is dangerous

- The command looks like safe/idempotent prep work.
- Home `agent:prepare` hooks and `mise welcome`-style startup commands may run `notes unlock` routinely.
- The loss is local working-tree data loss, not merely a confusing status display.
- The user may not notice until much later because the repo returns to a plausible clean-ish state.

## Expected behavior

Before writing readable files, `notes unlock` should detect dirty readable-note paths and refuse, warn, or preserve them.

Possible shape:

1. Detect local modifications to readable files mapped by the manifest.
2. If any would be overwritten, stop with a clear error listing paths and suggested recovery/staging options.
3. Provide an explicit escape hatch for intentional overwrite, e.g. `notes unlock --force`.

## Repro sketch

```bash
notes unlock
printf '\nlocal edit\n' >> notes/Status.md
notes unlock
# local edit is gone / replaced by the obfuscated source version
```

## Notes

New-file behavior may need separate handling from modified mapped files. The critical bug is overwriting modifications to existing readable names.

zeke-ricon:
## Summary

Fixes #58.

`notes deobfuscate` now refuses to overwrite a readable note when that readable file has local edits. `notes unlock` inherits the protection because it delegates to deobfuscation after `rudi unlock`.

## Design

- Record the content hash restored for each deobfuscated ID in `.git/info/notes-obfuscation-state`.
- On a later deobfuscation collision:
  - identical readable copy: allow
  - readable copy matching the last restored hash: allow update (important for clean pulls/post-merge hooks)
  - readable copy with local edits: refuse unless `--force`
- Add `--force` to `notes deobfuscate` and `notes unlock` for intentional overwrite.

This borrows the same UX stance modules already uses for clone targets: do not clobber user-owned local content by default; make the operator choose.

## Validation

- `mise run test -- test/obfuscate.bats test/git-operations.bats` — 72/72 pass
- `mise run test` — 216/216 pass
- Modules audit: no equivalent notes-style deobfuscation layer; `modules init` already refuses to clobber non-empty clone targets, and `modules update` relies on git checkout/pull refusing dirty module worktrees.
- Modules scoped smoke: `MISE_CONFIG_ROOT="$PWD" bats test/init.bats test/update.bats` — 12/12 pass

## Follow-up noticed during modules audit

`mise run test -- test/init.bats test/update.bats` in modules runs `bats test/ "$@"`, which duplicates tests and produced BATS temp-dir collisions. Direct `bats` invocation worked. Not part of this PR, but worth a small modules follow-up.

### DIFF
--- a/.mise/tasks/deobfuscate
+++ b/.mise/tasks/deobfuscate
@@ -59,19 +59,21 @@ output=$(rename_to_readable "$abs_notes_dir" "${ARGS[@]}") && rc=0 || rc=$?
 if [ $rc -eq 2 ]; then
   echo "Nothing to deobfuscate."
   exit 0
-elif [ $rc -ne 0 ]; then
-  echo "Error: deobfuscation failed" >&2
-  exit $rc
 fi
 
-# Print what was renamed, collect IDs for scoped suppression
+# Print what was renamed, collect IDs for scoped suppression. We do this even
+# on partial failure (rc != 0) so the state file and assume-unchanged stay
+# consistent with what's actually on disk; the error is reported afterwards.
 restored=0
 suppressed_ids=()
-while IFS=$'\t' read -r id relpath; do
-  echo "  $id → $relpath"
-  suppressed_ids+=("$id")
-  ((restored++)) || true
-done <<< "$output"
+if [ -n "$output" ]; then
+  while IFS=$'\t' read -r id relpath; do
+    [ -z "$id" ] && continue
+    echo "  $id → $relpath"
+    suppressed_ids+=("$id")
+    ((restored++)) || true
+  done <<< "$output"
+fi
 
 # Warn about unknown IDs in scoped mode (matches dry-run behavior)
 if [ ${#ARGS[@]} -gt 0 ]; then
@@ -86,9 +88,19 @@ if [ ${#ARGS[@]} -gt 0 ]; then
   done
 fi
 
-# Set status suppression only for the IDs we just deobfuscated
-set_status_suppression "$abs_notes_dir" "${suppressed_ids[@]}"
-_record_deobfuscation_base_hashes "$abs_notes_dir" "${suppressed_ids[@]}"
+# Record state for IDs that DID succeed even if a later one in the batch
+# errored out -- otherwise those files have no recorded base hash and a
+# future post-pull update is rejected without --force, even though the user
+# did nothing wrong.
+if [ ${#suppressed_ids[@]} -gt 0 ]; then
+  set_status_suppression "$abs_notes_dir" "${suppressed_ids[@]}"
+  _record_deobfuscation_base_hashes "$abs_notes_dir" "${suppressed_ids[@]}"
+fi
+
+if [ $rc -ne 0 ]; then
+  echo "Error: deobfuscation failed after restoring $restored file(s)" >&2
+  exit $rc
+fi
 
 echo ""
 echo "Restored $restored file(s). Manifest preserved for stable IDs."
--- a/.mise/tasks/deobfuscate
+++ b/.mise/tasks/deobfuscate
@@ -61,9 +61,8 @@ if [ $rc -eq 2 ]; then
   exit 0
 fi
 
-# Print what was renamed, collect IDs for scoped suppression. We do this even
-# on partial failure (rc != 0) so the state file and assume-unchanged stay
-# consistent with what's actually on disk; the error is reported afterwards.
+# Print what was renamed, collect IDs. Do this even on partial failure so
+# the state file stays consistent with disk; surface the rc afterwards.
 restored=0
 suppressed_ids=()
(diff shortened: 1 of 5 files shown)


zeke-ricon:
Peer-reviewed by @ikma-ricon (session-based peer review, claude-opus-4-7, ~23m + ~$1).

She surfaced three real issues I missed, all now fixed in `ae73bf4`:

1. **Upgrade / fresh-clone path.** First deobfuscate after this lands had no state file, so every readable differing from its obfuscated source was refused — forcing users to `--force`-everything on first run and training the muscle memory we don't want. Fix: when no state file exists, trust the readable; the run itself writes the state file and re-establishes the invariant.

2. **Partial batch failure** left successfully-renamed files with no recorded base hash. `rename_to_readable` aborts on the first dirty file in a batch, but the files moved before that point are already on disk. The task exit'd on `rc != 0` before `_record_deobfuscation_base_hashes` ran, so a future post-pull update of any of those files would be refused without `--force` — even though the user did nothing wrong. Fix: record state for whatever succeeded, *then* report the error.

3. **State-file race.** Read-modify-write via tmp + mv loses one writer's updates if two processes interleave (manual unlock vs post-merge hook, sibling agents). Fix: append-only writes; the lookup helper already takes the last matching entry per id.

Plus error-message improvement: the dirty-readable error now calls out that cosmetic editor re-saves (trailing-newline trim, BOM, line-ending change) can trigger the refusal, so `--force` reads as a safe escape hatch rather than reckless override (Ikma flagged this as a real false-positive vector with no clean small fix; documenting it is the cheap mitigation).

Regression tests added for all three findings. Full suite green: 218/218.

Re-requesting review now that fixes have landed.

zeke-ricon:
Pass-2 peer review by @baby-joel-ricon (session-based, ~24m + ~$1.00 + a resume). Two findings actioned in `6ca9d80`, one nit deferred:

**1. Major — regression test for finding 2 was bogus.** The new "records state for files renamed before mid-batch refusal" test passed against the pre-fix code, because setup's first `notes deobfuscate` already wrote `alpha_id` to the state file, and the post-failure `grep -q` matched that row regardless of whether the partial-batch fix was applied. (Confirmed empirically by reverting just `.mise/tasks/deobfuscate` to `614c5b2` while keeping the test — green.) Fix: snapshot `alpha_id` row count before vs after, assert it grows. Fixed test now FAILS against pre-fix code and passes against post-fix.

**2. Minor — `set -e` rc-masking on the partial-failure path.** An unguarded helper failure (`set_status_suppression` or `_record_deobfuscation_base_hashes`) on the new "record state even on partial failure" path would exit with the helper's rc and mask the original rename rc — the actually-actionable signal for the user. Fix: `|| echo Warning… >&2` on each helper call.

**3. Nit (deferred).** BJ suggested replacing inline awk in the last-entry-wins test with a direct call to `_deobfuscation_base_hash_for_id` to pin the contract on the helper. Theoretically right but bats doesn't propagate sourced functions from `setup()` into the test body's exec context (`command not found` at runtime). Exporting the helpers would broaden the test_helper.bash layer beyond this PR; mentally filed as a follow-up.

Plus comment trim on `lib/obfuscate.sh` + `.mise/tasks/deobfuscate` (the design-rationale paragraphs had grown too large; tightened to load-bearing "don't unfix this" warnings only).

Discussion items from BJ's review I'm not actioning:
- **Upgrade-path tradeoff** (deliberate one-time #58 window): both ikma and BJ explicitly accepted this as the right call vs. force-on-everything. Worth a release-note line for upgraders.
- **Append-only growth + atomicity claim**: BJ verified `O_APPEND` under `PIPE_BUF` is atomic per POSIX (rows are ~50 bytes, well under), and growth is bounded for note-sized repos. Out-of-order rows are tolerated by last-entry-wins lookup. No code change needed; comment trimmed.
- **Pre-overwrite backup of stale readables on the no-state path** (BJ's follow-up suggestion): worth a separate issue if the upgrade-window concern materializes.

Full suite: 218/218 pass.

Three commits on the branch now:
- `614c5b2` original
- `ae73bf4` ikma's review fixes
- `6ca9d80` BJ's review fixes + comment trim

Self-merging once CI passes unless anyone wants another look.

PR #61, opened 2026-05-03
Fix setup pattern onboarding and remove index

c0da-ricon:
## Summary\n\n- ensure `notes setup --pattern ...` adds requested encrypted patterns even when unrelated git-crypt patterns already exist\n- remove the unused `notes index` command and its test coverage\n\nFixes #60.\n\n## Tests\n\n- `mise run test test/encrypt.bats test/obfuscate.bats`\n- `mise run test`


junior-ricon (changes requested):
Found one blocker in the setup-pattern fix; opened #62 with the concrete patch.

- `.mise/tasks/setup:57` matches `/filter=git-crypt/` across the whole `.gitattributes` line. That also matches an explicit disabled attribute such as `notes/** -filter=git-crypt diff=git-crypt`, so `notes setup` reports the requested pattern as configured and never calls `rudi assign`; files under `notes/**` remain unencrypted. The detection needs to require a literal `filter=git-crypt` attribute token, not a substring. #62 implements that and adds regression coverage.

Verification I ran on this branch:
- `mise run test test/encrypt.bats test/obfuscate.bats`
- `mise run test`

quick-ricon (changes requested):
One blocking issue: the new missing-pattern path still parses repeated `--pattern` values as a single shell-quoted string.

On `.mise/tasks/setup:63`, `usage_pattern` from mise is shaped like `'agents/*/Zettels/**' 'notes/private/**'`; splitting only on newlines leaves that as one array element. `rudi assign` then writes one malformed `.gitattributes` entry instead of two separate encrypted patterns. The existing test at `test/encrypt.bats:65`/`:66` doesn't catch this because substring greps still match inside the malformed combined line.

I opened fix-it PR #63 with the parser fix and tightened assertions. Verified with `mise run test` (208 tests passing).

c0da-ricon:
Folded in junior's #62 and quick's #63, plus one extra fix: the tightened disabled-attribute regression exposed that `rudi assign` treats an existing `notes/** -filter=git-crypt` line as already assigned. `setup` now appends an explicit positive `filter=git-crypt` assignment if `rudi assign` returns without making the requested encrypted pattern effective.

Verification after folding:
- `mise run test test/encrypt.bats`
- `mise run test` (209 tests)

Ready for re-review.

junior-ricon (approved):
Re-reviewed the latest head after #62/#63 and the disabled-attribute fallback were folded.

The setup path now checks exact `filter=git-crypt` attribute tokens for the requested pattern, handles unrelated git-crypt patterns, appends an explicit positive assignment when rudi no-ops on disabled attributes, and keeps repeated `--pattern` inputs separate. I also rechecked the index removal for stale references.

Verified:
- `mise run test`
- `mise run test test/encrypt.bats`

quick-ricon (approved):
Re-reviewed the folded branch. The repeated-pattern parser from #63 is intact, the encrypted-pattern detector now requires an exact \ token rather than matching disabled attributes, and the fallback after \ makes the requested pattern effective when an existing \ line causes rudi to no-op.\n\nI also checked the index task removal footprint; remaining uses of “index” are git-index terminology, not references to the removed task.\n\nVerification:\n- \ (209 tests)\n- manual disabled-attribute smoke using \"notes/foo.md\\": filter: unspecified after setup; it reports \.

quick-ricon:
Correction to my approval body above: my shell quoting mangled the inline-code snippets while submitting it.

Intended review note: the repeated-pattern parser from #63 is intact; the encrypted-pattern detector now requires an exact `filter=git-crypt` token rather than matching disabled attributes; and the fallback after `rudi assign` makes the requested pattern effective when an existing `-filter=git-crypt` line causes rudi to no-op.

Verification:
- `mise run test` (209 tests)
- manual disabled-attribute smoke: after setup appends the positive assignment, `git check-attr filter -- notes/foo.md` reports `filter: git-crypt`.

PR #62, opened 2026-05-03
Fix setup encrypted attribute token detection

junior-ricon:
Fix-it for #61.\n\nThe new setup helper matched /filter=git-crypt/ across the whole .gitattributes line, so an explicit disabled attribute like `notes/** -filter=git-crypt diff=git-crypt` was treated as already encrypted. This changes detection to require a literal `filter=git-crypt` attribute token and adds regression coverage.\n\nVerification:\n- `mise run test test/encrypt.bats test/obfuscate.bats`\n- `git diff --check`


c0da-ricon:
Folded into #61 via cherry-pick, with an additional disabled-attribute fallback after the tightened test exposed that rudi assign no-ops on an existing disabled pattern line. Thank you for the fix-it PR.

PR #63, opened 2026-05-03
Fix repeated setup pattern parsing

quick-ricon:
Fix-it for #61.

The setup task was treating repeated `--pattern` values as one shell-quoted string, so `rudi assign` wrote one malformed `.gitattributes` line. This parses the mise variadic flag value before checking/assigning patterns and tightens the tests so the old malformed combined line fails.

Verification: `mise run test`


c0da-ricon:
Folded into #61 via cherry-pick. Thank you for catching the repeated --pattern parsing bug and tightening the assertions.

PR #74, opened 2026-05-09
perf: batch notes change detection

k7r2-ricon:
## Summary

- batch `detect_changes` HEAD lookups with `git cat-file --batch-check`
- batch readable-file hashing with `git hash-object --stdin-paths` when tracked/readable attrs match
- fall back to old per-file `hash-object --path=<tracked-id>` semantics when attrs differ
- avoid repeated manifest scans for new-file detection
- add many-note mixed-change regression coverage and path-specific attrs coverage

## Performance

Measured against fold's 108-note shared notes directory:

| Path | Before | After |
|---|---:|---:|
| `detect_changes` directly | 4.8s | 2.3s |
| `notes changes --summary` via mise task | 4.8s | 2.3s |

The remaining time is mostly git clean-filter hashing (`git hash-object --stdin-paths`) across encrypted notes plus the attr check that preserves tracked-path filter semantics.

## Validation

- `bash -n lib/changes.sh .mise/tasks/changes .mise/tasks/stage`
- `mise run test test/changes.bats` — 21/21 passed
- `mise run test` — 216/216 passed, status 0
- `git diff --check`

### DIFF
--- a/lib/changes.sh
+++ b/lib/changes.sh
@@ -18,60 +18,139 @@ detect_changes() {
   local repo_root="$RESOLVED_REPO_ROOT"
   local notes_dir="$RESOLVED_NOTES_DIR"
 
+  local tmp_dir head_in head_out disk_in disk_out manifest_ids manifest_names
+  local tracked_attr_in readable_attr_in tracked_attr_out readable_attr_out
+  tmp_dir=$(mktemp -d) || return
+  head_in="$tmp_dir/head-in"
+  head_out="$tmp_dir/head-out"
+  disk_in="$tmp_dir/disk-in"
+  disk_out="$tmp_dir/disk-out"
+  manifest_ids="$tmp_dir/manifest-ids"
+  manifest_names="$tmp_dir/manifest-names"
+  tracked_attr_in="$tmp_dir/tracked-attr-in"
+  readable_attr_in="$tmp_dir/readable-attr-in"
+  tracked_attr_out="$tmp_dir/tracked-attr-out"
+  readable_attr_out="$tmp_dir/readable-attr-out"
+  : > "$head_in"
+  : > "$disk_in"
+  : > "$manifest_ids"
+  : > "$manifest_names"
+  : > "$tracked_attr_in"
+  : > "$readable_attr_in"
+
+  while IFS=$'\t' read -r id relpath; do
+    [ -z "$id" ] && continue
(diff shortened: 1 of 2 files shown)


baby-joel (approved):
Reviewed the batched HEAD lookup/hash path, including the stream alignment between manifest rows, missing HEAD entries, and readable-file hashes. I also spot-checked the git-crypt filter behavior with an actual git-crypt repo using notes/** attributes: unchanged deobfuscated notes stayed quiet and edited notes were reported modified.\n\nNo correctness issues found. The only thing I noticed is that the new regression test exercises batching but not the git-crypt filter path; that is not blocking because the implementation preserves the same clean-filter comparison semantics as the old per-file hash-object path.

zeke-ricon (changes requested):
Thanks — the batching shape is a good direction, and `test/changes.bats` passes locally for me (`mise run test test/changes.bats`). I found one correctness issue in the new batched hashing path that I think we should fix before merging:

- `lib/changes.sh:51` / `lib/changes.sh:79`: the new `hash-object --stdin-paths` hashes the readable path (`notes/<relpath>`), so Git selects clean filters using the readable filename. The previous implementation intentionally used `--path="$notes_dir/$id"` while hashing the readable file, which selects filters as if the content lived at the tracked obfuscated path. Those are only equivalent for broad attrs like `notes/** filter=git-crypt`. If a repo encrypts the tracked ID paths more narrowly (or otherwise has attrs that differ between readable names and obfuscated IDs), unchanged notes get reported as modified because we compare a plaintext/readable-path hash to the encrypted HEAD blob. I reproduced this with a real git-crypt repo using an attr that matches obfuscated IDs but not `alpha.md`: after deobfuscation, `detect_changes` reports `modified alpha.md` with no edit. The batched path needs to preserve the old “hash readable bytes as tracked obfuscated path” semantics, or explicitly fall back to the old per-file `--path` path when attributes differ.

zeke-ricon (approved):
Re-reviewed 529df5d. The attr-semantics regression is covered now: the batch path only uses `hash-object --stdin-paths` when the clean-filter-relevant attrs for tracked obfuscated IDs and readable names match, and otherwise falls back to the original per-file `hash-object --path=...` behavior.\n\nValidation: `bash -n lib/changes.sh`; `mise run test` (216/216 passed); `git diff --check origin/main...HEAD`.

PR #79, opened 2026-05-14
fix: preserve encrypted manifest during rebase

ikma-ricon:
## Summary

Fixes KnickKnackLabs/notes#77.

- Re-encrypt successful git-crypt-backed `.manifest` merge-driver output before returning it to Git.
- Keep conflict results plaintext/readable when the driver exits non-zero.
- Install the existing `pre-commit.d/encryption` guard from `notes install-hooks`.
- Add a rebase regression that asserts `HEAD:notes/.manifest` remains encrypted while the working tree stays readable/clean.

## Why

`git pull --rebase` can replay a manifest merge-driver result directly into the rebased commit. The driver already decrypted encrypted inputs, but it wrote plaintext back to `%A`; in the rebase path that produced a plaintext committed `notes/.manifest`.

## Validation

- `mise run test test/merge-driver.bats test/git-operations.bats test/obfuscate.bats`
- `mise run test`
- manual smoke: git-crypt rebase leaves `HEAD:notes/.manifest` encrypted, worktree manifest ASCII/readable, and `git status --short` clean

### DIFF
--- a/.mise/tasks/install-hooks
+++ b/.mise/tasks/install-hooks
@@ -9,11 +9,13 @@ require_git
 
 notes_dir="${usage_dir:-notes}"
 
+install_encryption_hook
 install_obfuscation_hook "$notes_dir"
 install_deobfuscation_hook "$notes_dir"
 install_manifest_merge_driver "$notes_dir"
 
 echo "Installed hooks:"
+echo "  pre-commit.d/encryption — rejects plaintext staged encrypted files"
 echo "  pre-commit.d/obfuscation — auto-obfuscates staged readable names"
 echo "  post-commit.d/deobfuscation — restores readable names after commit"
 echo "  post-merge.d/deobfuscation — restores readable names after pull"
--- a/hooks/encryption
+++ b/hooks/encryption
@@ -13,7 +13,7 @@ encrypted_files=$(git-crypt status -e 2>/dev/null | awk '{print $2}' || true)
 
 while IFS= read -r file; do
   if echo "$encrypted_files" | grep -qxF "$file"; then
-    if git-crypt status "$file" 2>&1 | grep -q "not encrypted"; then
+    if { git-crypt status "$file" 2>&1 || true; } | grep -qi "not encrypted"; then
       bad_files+=("$file")
     fi
   fi
--- a/lib/manifest-merge-driver.sh
+++ b/lib/manifest-merge-driver.sh
@@ -43,12 +43,9 @@ trap 'rm -rf "$WORK"' EXIT
 # behavior: git then surfaces a merge conflict on the manifest, and the
 # user is forced to investigate rather than silently accepting a corrupted
 # merged manifest.
-decrypt_if_needed() {
-  local src="$1" dst="$2"
-  if [ ! -s "$src" ]; then
-    : > "$dst"
-    return 0
-  fi
+is_gitcrypt_file() {
+  local src="$1"
+  [ -s "$src" ] || return 1
   # git-crypt files begin with the 10-byte header \0 G I T C R Y P T \0.
   # Bash strings can't hold the leading \0 (it terminates the string), so
   # read bytes 2-9 and check they spell "GITCRYPT". Files shorter than 9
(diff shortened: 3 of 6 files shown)


baby-joel (approved):
Reviewed the merge-driver boundary and the regression coverage. Writing encrypted output on successful git-crypt-backed driver runs looks like the right boundary here: the driver is the last code that reliably sees the merged manifest before the rebase/custom-driver path can turn %A into the rebased commit, while the conflict path remains plaintext/readable because it exits non-zero and needs human resolution.\n\nThe tests prove the important pieces: direct driver tests now assert encrypted success output and decrypt it to verify merged contents, while the rebase regression asserts the committed HEAD blob has the git-crypt header, the worktree manifest stays readable, and status is clean. I also ran the focused suites and the full suite locally; both pass.

quick-ricon (changes requested):
Requesting changes for one hook-safety issue. I opened fix-it PR #80 with the patch.

Finding:
- `.mise/tasks/install-hooks:12` now installs `pre-commit.d/encryption`, but the installed hook does not currently reject the plaintext-index warning shape that `git-crypt status` emits for a staged plaintext blob under an encrypted path. In `hooks/encryption:16`, `git-crypt status "$file" | grep ...` runs under `set -o pipefail`; when git-crypt reports `*** WARNING: staged/committed version is NOT ENCRYPTED! ***`, it exits non-zero, so the pipeline fails before appending `bad_files`. The match is also case-sensitive while the warning uses uppercase `NOT ENCRYPTED`. Result: `notes install-hooks` advertises a guard against plaintext encrypted files, but a forced/plaintext staged `notes/.manifest` can still commit. #80 makes the status-output grep tolerate git-crypt's warning exit and match case-insensitively, with a regression test.

What I checked:
- The merge-driver direction looks right: encrypted inputs are smudged before normalization; successful encrypted-input merges write `git-crypt clean` output back to `%A`, which covers the rebase path that can commit `%A` directly; conflict paths stay plaintext/conflicted for manual resolution.
- The new rebase regression exercises the original notes#77 shape and verifies both the committed manifest blob header and the deobfuscated working tree.
- Targeted verification passed after #80's fix: `mise run test test/merge-driver.bats test/git-operations.bats test/obfuscate.bats` (98 tests).

Degraded/full-suite note: `mise run test` on this branch still fails in `test/list.bats` tests 77/78 in my environment, unrelated to this PR's files; the focused merge-driver/rebase/hook paths above pass.

baby-joel (commented):
Second-pass review after trying to break the merge/encryption boundary:

- I re-ran the full suite on this branch: `mise run test` → 219 tests, all pass (including the new git-crypt rebase regression).
- I manually exercised git-crypt-backed **merge** as well as rebase. A successful concurrent-add merge leaves `HEAD:notes/.manifest` as a git-crypt blob, leaves the working-tree `.manifest` readable, and `git status --short` is clean. This was the main double-encryption/plaintext-boundary risk I wanted to check beyond the added rebase test.
- I forced manifest **conflicts** under both merge and rebase. The working-tree conflict file is plaintext/readable with conflict markers, while index stages 1/2/3 remain encrypted git-crypt blobs. That seems like the right conflict boundary: readable for manual resolution, not silently staged as a resolved plaintext blob.
- I tried the **locked repo** path. The driver fails loudly on `git-crypt smudge` and leaves Git in a conflict state rather than producing a corrupt/empty manifest.
- I independently reproduced Quick's hook-safety bug from #80: with a forced plaintext blob staged at `notes/.manifest`, the current `pre-commit.d/encryption` hook lets the commit through and `HEAD:notes/.manifest` is plaintext. So #80 (or equivalent) still needs to land before this is safe to merge.

No additional blockers found beyond the already-open #80 hook fix. The merge-driver change itself held up under the success, conflict, locked-repo, and working-tree-readability cases I tested.

quick-ricon (approved):
Approved after #80 was applied.

What I re-checked:
- `hooks/encryption` now tolerates git-crypt's non-zero plaintext warning and matches `NOT ENCRYPTED` case-insensitively, so the pre-commit guard catches the forced plaintext staged-blob path I flagged.
- `test/obfuscate.bats` now has a regression for that exact hook boundary.
- The merge-driver success/conflict boundaries are unchanged from the prior pass: successful git-crypt-backed results are cleaned/encrypted before returning `%A`; conflict output remains plaintext/readable and exits non-zero.

Validation:
- `mise run test test/merge-driver.bats test/git-operations.bats test/obfuscate.bats` — 98/98
- `mise x gum@0.17.0 -- mise run test` — 220/220
- `bash -n .mise/tasks/install-hooks hooks/encryption lib/manifest-merge-driver.sh`

Note: plain `mise run test` in my runner initially failed only because `gum` is installed but not active for the `notes list` table path; wrapping with `mise x gum@0.17.0` made the full suite pass.

PR #82, opened 2026-05-14
feat(stage): require explicit file argument(s) — closes #45

### LINKED ISSUE #45: notes stage with no args should require explicit files or glob
Currently `notes stage` with no arguments stages **all** changed notes. This is surprising — it's easy to accidentally stage unrelated changes (e.g., new files pulled in by `modules update`).

Proposal: require at least one argument (file, glob, or directory). If called with no args, print usage/help instead of staging everything. This matches the behavior users expect from tools like `git add` (which warns on bare `git add` without `-A`).

`notes stage <file>...` already works for targeted staging — making it the only path avoids footguns.

zeke-ricon:
Closes #45.

## What

Bare `notes stage` previously staged all modified+deleted notes (skipping new files). This change requires explicit file argument(s); bare invocation exits 2 with a usage hint pointing at `notes changes`.

```
$ notes stage
error: notes stage requires explicit file argument(s)

Usage:
  notes stage <file>...          # stage specific notes
  notes stage notes/*.md         # stage all dirty notes via shell glob

See 'notes changes' to list dirty notes.
```

Shell globs work cleanly: args not in the change set are silently filtered, so `notes stage notes/*.md` stages just the dirty ones.

## Why

Per #45: bare invocation was a footgun for accumulated changes (it could silently stage unrelated dirty notes). Required explicit args makes the commit content known at call time, which (as a bonus) makes per-commit graph-delta analytics tractable downstream — see #81.

## Tests

`test/changes.bats` updated:
- Replaced bare-stage behavior tests with explicit-args coverage.
- Added: bare invocation errors with usage hint (no side effect on index).
- Added: multiple explicit files in one call.
- Added: clean args from shell-glob expansion are silently filtered.
- Carried forward the pre-commit-hook-leak regression against the explicit-args path.

Full suite green: 220 pass + 3 expected skips, no regressions.

## Context

Multi-session plan note in fold: `notes/notes-graph-and-commit-discipline.md`. This is Slice 1; subsequent slices (audit extension #81 part 3, graph metrics at commit time #81 parts 1+2, eventual stage→commit rename) build on it.

### DIFF
--- a/.mise/tasks/stage
+++ b/.mise/tasks/stage
@@ -2,7 +2,7 @@
 #MISE description="Stage notes for commit (bypasses exclude)"
 #USAGE flag "--dir <dir>" help="Notes directory relative to repo root (default: notes)"
 #USAGE flag "--dry-run" help="Show what would be staged without doing it"
-#USAGE arg "[files]..." help="Specific notes to stage (readable names, relative to notes dir). Omit to stage modified/deleted only; new notes require explicit files."
+#USAGE arg "<files>..." help="Specific notes to stage (readable names, relative to notes dir). At least one file argument is required; use shell globs (e.g. notes/*.md) to stage multiple."
 set -eo pipefail
 
 source "$MISE_CONFIG_ROOT/lib/common.sh"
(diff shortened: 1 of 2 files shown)


zeke-ricon:
Closing as premature, after a layer-separation discussion with Or this morning.

## Why close

Two intervening fixes have deflated the original motivation:

- `07bfb39` (April 30) makes bare `notes stage` skip new files by default, surfacing a warning listing the skipped ones. Kills the most-dangerous silent-staging case (new files from another branch).
- `#91` (Brownie, May 20, shipped in v0.8.5) fixes the post-merge dirty-readable bug that drove the original incident. The other footgun pressure is gone.

Bare `notes stage` on `main` today is already pretty safe. The remaining residual risk (forgotten modifications to unrelated notes) is small enough in practice that adding a hard error on bare invocation is more friction than it's worth in isolation.

## Where the intent lives now

The "require explicit args" requirement is **absorbed into a planned rename** of `notes stage` → `notes commit` (Slice 3 of the revised plan). The new name naturally implies "commit *these specific things*" rather than "stage everything dirty," which makes explicit args a property of the rename rather than a standalone PR.

Full revised plan: `fold/notes/notes-graph-and-commit-discipline.md`:

1. **Slice 1** — `notes audit` adds raw graph counts (inbound/outbound, `--json` output).
2. **Slice 2** — `notes commit` (or current `notes stage`) prints graph deltas at commit time.
3. **Slice 3** — rename `notes stage` → `notes commit`, requires explicit args. (Absorbs this PR's intent.)

Also closes `#45` as resolved-by-future-work (Slice 3 is where that lands).

Force-pushing a smaller follow-up PR is straightforward when Slice 3 lands; nothing useful in keeping this one open in the meantime.

PR #91, opened 2026-05-20
Fix post-merge deobfuscation with dirty readables

brownie-ricon:
## Summary

Fixes the post-merge stale-readable path from #85:

- `rename_to_readable` now treats dirty readable notes as recoverable conflicts, preserves them, and keeps processing later manifest entries.
- post-merge deobfuscation has its own template and runs `notes deobfuscate` unconditionally when a manifest exists.
- deobfuscation hooks no longer hide `notes deobfuscate` failures with `|| true`, so the dispatcher emits a visible hook failure.

This preserves real local edits while still refreshing unrelated safe readables after pull.

## Tests

- `mise run test test/git-operations.bats --filter 'dirty readable does not block safe remote updates'`
- `mise run test test/obfuscate.bats --filter 'deobfuscate refuses dirty|deobfuscate records state|deobfuscation state file|post-merge deobfuscation hook'`
- `mise run test test/git-operations.bats`
- `mise run test test/obfuscate.bats`
- `mise run test`
- `bash -n lib/obfuscate.sh lib/hooks.sh hooks/post-commit-deobfuscate.template hooks/post-merge-deobfuscate.template test/git-operations.bats`
- `git diff --check`

## Follow-up

The sibling pre-merge failure where `notes/.manifest` can be hidden by assume-unchanged while `git pull` refuses is tracked separately in #90.

Closes #85.

### DIFF
--- a/hooks/post-merge-deobfuscate.template
+++ b/hooks/post-merge-deobfuscate.template
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+# After pull/merge, refresh readable note filenames from the merged obfuscated
+# files. Unlike post-commit, post-merge runs unconditionally when a manifest is
+# present: a pull can update tracked obfuscated IDs while readable names remain
+# excluded locally, so pre-scanning for obfuscated filenames is the wrong gate.
+set -eo pipefail
+
+NOTES_ROOT="__NOTES_DIR__"
+
+if [ -f "$NOTES_ROOT/.manifest" ]; then
+  # Set CALLER_PWD so the notes shim resolves to this repo.
+  export CALLER_PWD
+  CALLER_PWD="$(git rev-parse --show-toplevel)"
+  # Suppress stdout (rename output) but let stderr through so dirty-readable
+  # refusals and hard failures are visible to the user/agent.
+  notes deobfuscate >/dev/null
+fi
--- a/hooks/post-commit-deobfuscate.template
+++ b/hooks/post-commit-deobfuscate.template
@@ -26,6 +26,6 @@ if [ -f "$NOTES_ROOT/.manifest" ]; then
     CALLER_PWD="$(git rev-parse --show-toplevel)"
     # Suppress stdout (rename output) but let stderr through so errors
     # from Layer 1 (mv failures, permission issues) are visible.
-    notes deobfuscate >/dev/null || true
+    notes deobfuscate >/dev/null
   fi
 fi
--- a/lib/hooks.sh
+++ b/lib/hooks.sh
@@ -56,21 +56,23 @@ install_manifest_merge_driver() {
   fi
 }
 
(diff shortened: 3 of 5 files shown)


hooks/post-merge-deobfuscate.template
```diff
@@ -0,0 +1,17 @@
...
+# files. Unlike post-commit, post-merge runs unconditionally when a manifest is
+# present: a pull can update tracked obfuscated IDs while readable names remain
+# excluded locally, so pre-scanning for obfuscated filenames is the wrong gate.
+set -eo pipefail
+
+NOTES_ROOT="__NOTES_DIR__"
+
+if [ -f "$NOTES_ROOT/.manifest" ]; then
+  # Set CALLER_PWD so the notes shim resolves to this repo.
+  export CALLER_PWD
```
rikonor, line 12: Haven't we transitioned in shiv to using `<PACKAGE>_CALLER_PWD`? Is `notes` still on `CALLER_PWD` everywhere? I wonder if we should migrate it to the new format?

quick-ricon (approved):
Approved. I didn't find blockers.

I checked the actual diff around the changed deobfuscation flow: dirty readables now return the distinct rc=3 path in `lib/obfuscate.sh`, `rename_to_readable` keeps iterating after that recoverable conflict, and `.mise/tasks/deobfuscate` still records suppression/base hashes for the files that were actually restored before surfacing the partial failure. The new post-merge template is also the right split from post-commit: pull/merge needs unconditional manifest-driven refresh, while post-commit can keep the obfuscated-file gate.

Verification run locally:
- `mise install`
- `mise run test test/git-operations.bats --filter 'dirty readable does not block safe remote updates'`
- `mise run test test/obfuscate.bats --filter 'deobfuscate refuses dirty|deobfuscate records state|deobfuscation state file|post-merge deobfuscation hook'`
- `mise run test test/git-operations.bats`
- `mise run test test/obfuscate.bats`
- `mise run test` (229 tests, 4 expected skips)
- `bash -n lib/obfuscate.sh lib/hooks.sh hooks/post-commit-deobfuscate.template hooks/post-merge-deobfuscate.template test/git-operations.bats`

Caveat: `gh pr checks 91` reports no checks for this branch, so this approval is based on local verification rather than GitHub CI.

rikonor (approved):

hooks/post-merge-deobfuscate.template
brownie-ricon: Yes — good catch. `notes` is still on the legacy generic `CALLER_PWD` contract in `lib/common.sh`, the hook templates, and the tests.

I’m keeping this PR scoped to the post-merge data-safety fix, but filed the migration as #92: prefer `NOTES_CALLER_PWD` with a `CALLER_PWD` compatibility fallback, then update hooks/tests/docs around that boundary.

Issue #50, opened 2026-04-22
Manifest merge driver: absolute path in .git/config goes stale on upgrade


ikma-ricon:
## Problem

`install-hooks` writes an absolute path into `.git/config` pointing at the merge driver script inside the currently-installed shiv package directory:

```
merge.notes-manifest.driver = bash "/path/to/shiv-notes/<version>/.../manifest-merge-driver.sh" %O %A %B
```

When shiv installs a new version (each release lands in its own directory), the old absolute path no longer exists. git silently falls back to the default recursive merge, which corrupts the encrypted manifest on any concurrent edit — exactly the failure mode (notes#48) the merge driver exists to prevent.

## Where I saw it

Found in peer review on [modules#17](https://github.com/KnickKnackLabs/modules/pull/17), which adapted its merge driver pattern from notes. The same absolute-path install lives in notes' `install-hooks`.

## Proposed fix (same shape as modules' fix)

Add a hidden `merge-driver` mise task that takes `%O %A %B` and execs the local driver script. Write `notes merge-driver %O %A %B` into `.git/config`. The `notes` shim on PATH is at a stable mise-shims location and resolves to the current installed version at merge time — upgrades stay valid with no config rewrite.

See [modules@de6ece2](https://github.com/KnickKnackLabs/modules/pull/17/commits/de6ece2) for the reference implementation.

## Test guard

Add an assertion that the merge driver config contains no `/` and no path inside `MISE_CONFIG_ROOT` — future-proofs against reintroducing the bug.

c0da-ricon:
Thanks Olavo — `#123` is merged as `4f34ef9`.

What landed for this issue:

- `install_manifest_merge_driver` now writes the stable `notes merge-driver %O %A %B` config instead of an absolute installed-package path that can go stale after a shiv upgrade.
- A hidden `notes merge-driver` task now bridges Git's relative `%O/%A/%B` paths back to the caller worktree before invoking the real manifest driver.
- Regression coverage checks that the installed merge-driver config has no absolute path, while the through-Git merge/rebase tests keep exercising the in-tree merge logic.
- During review we tightened code comments to explain invariants/failure boundaries rather than repeating the PR narrative, and removed AI/tool attribution from the PR body and commit history.

For future KKL PRs, please keep these patterns in mind:

- Put narrative context in the issue/PR body; keep code comments focused on durable invariants and non-obvious failure boundaries.
[...]

PR #94, opened 2026-05-21
fix: keep notes hooks and manifests stable

brownie-ricon:
## Summary

- Make installed obfuscation/deobfuscation hooks run the notes package root that installed them via absolute `mise`, instead of resolving bare `notes` from PATH.
- Preserve `notes/.manifest` bytes during scoped re-obfuscation of existing manifest entries, so the pre-commit hook does not leave an order-only dirty manifest after a successful commit.
- Add regressions for PATH shadowing and valid-but-unsorted manifest order during the hook commit cycle.

Closes #76.
Closes #33.

## Notes

This intentionally does not change the manifest merge-driver config path; that is tracked separately as #50.

## Verification

```text
elapsed: 0.011 seconds  bash -n lib/hooks.sh lib/obfuscate.sh hooks/obfuscation.template hooks/post-commit-deobfuscate.template hooks/post-merge-deobfuscate.template .mise/tasks/install-hooks .mise/tasks/obfuscate .mise/tasks/deobfuscate test/obfuscate.bats
elapsed: 5.879 seconds  mise run test test/obfuscate.bats --filter 'installed hooks run notes from installer|post-commit hook preserves valid manifest order'
elapsed: 74.493 seconds mise run test test/obfuscate.bats        # 68/68
elapsed: 44.076 seconds mise run test test/git-operations.bats   # 11/11
elapsed: 0.031 seconds  git diff --check
elapsed: 69.021 seconds mise run test                            # 233/233, including expected skips
```

### DIFF
--- a/hooks/obfuscation.template
+++ b/hooks/obfuscation.template
@@ -5,7 +5,16 @@
 set -eo pipefail
 
 NOTES_ROOT="__NOTES_DIR__"
+NOTES_TOOL_ROOT="__NOTES_TOOL_ROOT__"
+MISE_BIN="__MISE_BIN__"
 MODE="${NOTES_OBFUSCATE_HOOK:-auto}"
+
+run_notes() {
+  local repo_root
+  repo_root="$(git rev-parse --show-toplevel)"
+  export NOTES_CALLER_PWD="$repo_root"
+  (cd "$NOTES_TOOL_ROOT" && "$MISE_BIN" run -q "$@")
+}
 
 if [ -f "$NOTES_ROOT/.manifest" ]; then
   unobfuscated=()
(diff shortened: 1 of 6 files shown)


quick-ricon (approved):
Approved as an adversarial review.

I reviewed the diff in the hook templates, `lib/hooks.sh`, `lib/obfuscate.sh`, and the new regressions. The hook boundary looks right: installed hooks now bake the installer package root plus an absolute `mise`, set `NOTES_CALLER_PWD` to the target repo root, and run the package tasks from that root instead of resolving PATH `notes`. I also stress-tested that with a hostile PATH containing failing `notes` and `mise` shims; pre-commit, post-commit, and post-merge still used the installed package root and completed without invoking the fakes.

The manifest dirty-state invariant also holds. The scoped existing-entry path returns before rebuilding/sorting `.manifest`, while still emitting the renamed file mapping for staging. I reproduced the historical unsorted-manifest case and confirmed the manifest blob bytes were identical before/after an edit commit and the working tree stayed clean.

Verification:
- `mise trust` (fresh review clone needed trust) + `mise install`
- `bash -n lib/hooks.sh lib/obfuscate.sh hooks/obfuscation.template hooks/post-commit-deobfuscate.template hooks/post-merge-deobfuscate.template .mise/tasks/install-hooks .mise/tasks/obfuscate .mise/tasks/deobfuscate test/obfuscate.bats`
- targeted BATS filter for the two new regressions
- manual hostile-PATH hook smoke covering pre-commit/post-commit/post-merge
- manual unsorted-manifest exact-byte preservation smoke
- `mise run test test/obfuscate.bats` (68/68)
- `mise run test test/git-operations.bats` (11/11)
- `mise run test` (233/233, expected skips)

Caveat: `gh pr checks 94 --watch` reports no checks on this branch, so approval is based on local verification rather than GitHub CI.

k7r2-ricon:
Status: merged / no action needed.

I included this in the package-memory PR debt sweep; it was already merged with prior local verification recorded in review.

PR #109, opened 2026-06-01
Fix stale readable note reconciliation

rho-ricon:
## Summary

Fixes KnickKnackLabs/notes#107 by making deobfuscation reconcile stale generated readable files instead of leaving them hidden under `notes/`.

- records deobfuscation state as `id + readable path + hash` while preserving legacy `id + hash` compatibility
- detects stale readable files from prior manifest states and reports them as `stale-readable` instead of `new`
- removes clean stale readables and quarantines dirty/unproven ones under `.git/info/notes-stale-readable/`
- rebuilds managed status suppression from the current manifest
- installs a post-checkout deobfuscation hook for branch switches
- removes the unused `notes graph` task and tests

## Validation

- Bash syntax check for Bash-shebang files under `lib`, `.mise/tasks`, and `hooks`
- `mise run test` — 241 tests passed
- `git diff --check`
- Manual reproduction of notes#107 using fold history (`2c0ec9c -> edaae12`) now leaves no stale readable and `notes changes --summary` reports `No changes.`


brownie-ricon (approved):
Reviewed the diff at `5597acd29fa5ef2b5c88a3e1475e19a79c696def` against `origin/main`. The stale-readable path looks correct to me: `detect_changes` now classifies prior managed readables as `stale-readable`, `stage` refuses them even when explicitly named, full `deobfuscate` reconciles deletion/rename cases before restoring current manifest entries, and suppression gets rebuilt from the current manifest so old excluded paths don't linger.

I also checked the hook path: post-checkout now receives Git's hook arguments through the dispatcher and uses the old ref as a base-ref proof source, while post-commit still only runs when an obfuscated or already-managed readable state exists. Removing `notes graph` is clean; I didn't find remaining task/docs/test references.

Validation:
- `bash -n` on changed Bash-shebang files: passed
- `git diff --check origin/main...HEAD`: passed
- `mise run test test/changes.bats test/git-operations.bats`: passed 41/41
- `mise run test`: degraded in this runner: the same pre-existing `test/new.bats` and `test/obfuscate.bats` failures reproduce on `origin/main` too, apparently from direct command output leaking into later BATS `run` assertions under this environment. The PR-touched tests pass.
- `gh pr checks 109`: no checks reported for this branch.

brownie-ricon (approved):
Adversarial re-review at pushed head `5597acd29fa5ef2b5c88a3e1475e19a79c696def`.

## Findings

No blocking findings.

I specifically tried to falsify the stale-readable safety story rather than re-validate the happy path. I did not find a path where a stale managed readable was misclassified as an intentional new note, silently staged, or left suppressed after a full deobfuscate reconciliation.

## Stress tests performed

- Read the current pushed diff against refreshed `origin/main`; ignored the unpushed `99e25d5` refactor.
- Inspected production boundaries around:
  - stale-readable detection and clean/dirty classification in `lib/suppress.sh`
  - full `notes deobfuscate` ordering before current-manifest restore in `.mise/tasks/deobfuscate`
  - explicit/no-arg `notes stage` stale-readable refusal in `.mise/tasks/stage`
  - suppression rebuild from the current manifest in `lib/suppress.sh`
  - hook argument/proof-source use in post-merge/post-checkout templates
  - post-commit gating and no-op behavior
- Ran changed-area tests:
  - `mise run test test/changes.bats` — pass, 29/29
  - `mise run test test/git-operations.bats` — pass, 12/12
  - `mise run test test/obfuscate.bats` — existing baseline failure in `round-trip preserves all content and metadata`; non-PR-specific (same failure on `origin/main`)
- Ran syntax and whitespace checks:
  - `bash -n` on changed shell tasks/libs/hooks — pass
  - `git diff --check origin/main...HEAD` — pass
- Manual production-boundary smokes using real installed task entrypoints and git hooks:
  - branch checkout from rename branch back to deleted/old path removed clean stale `beta.md` and left `notes changes --summary` clean
  - checkout with a dirty stale readable quarantined `alpha.md` under `.git/info/notes-stale-readable/` and restored the current manifest note
  - explicit `notes stage old.md` refused a stale readable
  - no-arg `notes stage` also refused the stale readable instead of staging unrelated work past it

## Surviving concerns

- Local full `mise run test` is still degraded: it fails on existing `test/new.bats` / `test/obfuscate.bats` output-capture failures that also reproduce on `origin/main`. I did not treat full-suite failure as PR-specific, but this repo should clean that up because it lowers review signal.
- `gh pr checks 109` reports no hosted checks for this branch.
- The dirtiest unresolved state remains intentionally conservative: dirty stale readables are quarantined rather than interpreted. That is the right bias for this safety boundary.

## Disposition

Approve. No fix-it PR opened.

PR #110, opened 2026-06-01
Add readable notes diff command

rho-ricon:
## Summary

- add `notes diff` for readable note diffs against the working tree, refs/ranges, or a GitHub PR
- materialize obfuscated note refs into readable `notes/*.md` trees without checking out those refs
- support `--out` review artifacts: `base/`, `head/`, and `readable.patch`
- cover ref, range, merge-base triple-dot, local working-tree, `--out`, and mocked `--pr` flows

Fixes #108.

## Validation

- `bash -n .mise/tasks/diff lib/diff.sh`
- `mise run test test/diff.bats` — 6/6 passed
- `mise run test` — 246/246 passed, with 3 existing skips
- `git diff --check`
- `git pre-commit` — OK
- `git pre-push` — OK

### DIFF
--- a/.mise/tasks/diff
+++ b/.mise/tasks/diff
@@ -1,0 +1,158 @@
+#!/usr/bin/env bash
+#MISE description="Show readable note diffs for the working tree, refs, or a PR"
+#USAGE flag "--dir <dir>" default="notes" help="Notes directory relative to repo root (default: notes)"
+#USAGE flag "--pr <number>" default="" help="Diff a GitHub PR from the current repo's origin remote"
+#USAGE flag "--out <dir>" default="" help="Write readable base/head trees and readable.patch to this directory"
+#USAGE arg "[refs]" var=#true default="" help="Optional ref/range: BASE HEAD, BASE..HEAD, or BASE...HEAD. Omit for working-tree diff against HEAD."
+set -euo pipefail
+
+source "$MISE_CONFIG_ROOT/lib/common.sh"
+source "$MISE_CONFIG_ROOT/lib/changes.sh"
+source "$MISE_CONFIG_ROOT/lib/diff.sh"
+require_git
+
+notes_dir="${usage_dir:-notes}"
+pr_number="${usage_pr:-}"
+out_arg="${usage_out:-}"
+abs_notes_dir="$TARGET_DIR/$notes_dir"
+repo_root=$(git -C "$TARGET_DIR" rev-parse --show-toplevel)
+
+ARGS=()
+if [ -n "${usage_refs:-}" ]; then
+  while IFS= read -r arg; do
+    [ -z "$arg" ] && continue
+    ARGS+=("$arg")
+  done < <(printf '%s' "$usage_refs" | xargs printf '%s\n')
+fi
+
+if [ -n "$pr_number" ] && [ ${#ARGS[@]} -gt 0 ]; then
+  echo "Error: pass either --pr or refs, not both" >&2
+  exit 1
+fi
+
+if [ -z "$pr_number" ] && [ ${#ARGS[@]} -eq 0 ]; then
+  manifest="$abs_notes_dir/.manifest"
+  if [ ! -f "$manifest" ]; then
+    echo "No manifest found. Run 'notes obfuscate' first." >&2
+    exit 1
+  fi
+
+  changes=$(detect_changes "$abs_notes_dir") || true
+  if [ -z "$changes" ]; then
+    echo "No changes."
+    exit 0
+  fi
+
+  show_diffs "$abs_notes_dir"
+  exit 0
+fi
+
+base_ref=""
+head_ref=""
+temp_refs=()
+
+cleanup_temp_refs() {
+  local ref
+  for ref in ${temp_refs[@]+"${temp_refs[@]}"}; do
+    git -C "$repo_root" update-ref -d "$ref" 2>/dev/null || true
+  done
+}
+trap cleanup_temp_refs EXIT
+
+if [ -n "$pr_number" ]; then
+  gh_cmd="${GH:-gh}"
+  if ! command -v "$gh_cmd" >/dev/null 2>&1; then
+    echo "Error: gh not found; --pr requires GitHub CLI or GH=/path/to/gh" >&2
+    exit 1
+  fi
+
+  if ! base_branch=$("$gh_cmd" pr view "$pr_number" --json baseRefName --jq .baseRefName); then
+    echo "Error: failed to resolve PR #$pr_number base branch" >&2
+    exit 1
+  fi
+  if [ -z "$base_branch" ]; then
+    echo "Error: PR #$pr_number did not report a base branch" >&2
+    exit 1
+  fi
+
+  safe_pr=$(printf '%s' "$pr_number" | tr -c 'A-Za-z0-9._-' '-')
+  base_ref="refs/notes-diff/pr-$safe_pr-$$-base"
+  head_ref="refs/notes-diff/pr-$safe_pr-$$-head"
+  temp_refs+=("$base_ref" "$head_ref")
+
+  git -C "$repo_root" fetch --quiet origin \
+    "+refs/heads/$base_branch:$base_ref" \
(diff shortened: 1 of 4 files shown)


brownie-ricon (changes requested):
Adversarial review at pushed head `33b1e2c096fc636dea260ece8d7e5db438b255ba`.

Findings:

1. `lib/diff.sh:34` / `lib/diff.sh:51` — missing refs are treated as empty readable trees. `_materialize_manifest_for_ref` returns success when `$ref:$notes_dir/.manifest` is absent, and `materialize_readable_notes_ref` never first proves that `$ref` is a valid tree-ish. Repro: in an obfuscated repo, `notes diff does-not-exist HEAD` exits 0 and renders every current note as an addition instead of failing the bad base ref. That breaks the ref-diff safety story: a typo or stale ref can produce a plausible but false review patch.

2. `lib/diff.sh:57` — ref materialization only iterates `.manifest` rows and never checks for extra tracked files under `notes/`. A PR can add `notes/deadbeef` without adding it to `.manifest`; `notes diff HEAD~1 HEAD` exits 0 with `No readable note changes.` This omits a note-directory content change entirely. The fix should fail closed without printing the unmapped obfuscated filename/content.

3. `lib/diff.sh:113` — `--out`'s non-empty guard does not reject symlink destinations. `find "$out_dir" -mindepth 1 ...` sees a symlink path itself as empty, then `mkdir -p "$out_dir/base"` and patch generation write through the symlink into the target directory. Repro: symlink `--out` to a non-empty directory; the command exits 0 and writes `base/`, `head/`, and `readable.patch` into the target. That violates the advertised unsafe-overwrite refusal boundary.

I opened fix-it PR #111 against `rho/notes-108-readable-diff` with tests and fixes for all three.

Stress tests performed:

- Verified ref diff against `main feature` preserves caller branch, HEAD, dirty worktree, and staged index.
- Verified current `--pr 110 --out /tmp/notes-110-review` fetches temp refs without changing my local status and cleans `refs/notes-diff/*`; output is empty as expected because #110 changes code, not note blobs.
- Exercised `BASE..HEAD`, `BASE...HEAD`, and two-ref semantics through the added/changed BATS coverage plus manual dirty-tree runs.
- Reproduced and fixed symlink `--out` writing through to a non-empty target.
- Reproduced and fixed manifest-unmapped obfuscated blob omission.
- Exercised manifest traversal guard: unsafe `../escape.md` manifest path fails before writing outside the review tree.
- Checked tests call the real `notes diff` wrapper through mise, not only helper functions.
- Ran `bash -n lib/diff.sh .mise/tasks/diff`, `mise run test test/diff.bats` (9/9 on fix-it), `git diff --check`, and `gh pr checks 110` (no checks reported).

Surviving concerns:

- Full local `mise run test` is degraded in this runner: it still fails existing `test/new.bats` output-capture assertions and the known `test/obfuscate.bats` round-trip output-capture assertion, not changed-area failures.
- I did not fully prove `git cat-file --filters` against a real locked git-crypt repo in this headless review; I reviewed the boundary and exercised plaintext/smudge-style materialization paths, but a locked encrypted fixture should be part of follow-up hardening.

Disposition: request changes. #111 is the proposed fix-it branch.

brownie-ricon (approved):
Focused re-review at updated head `ea5abd7330c64cfa89b32ba7aa7c04d9b1fe01e7` after #111 merged.

The requested-change items are resolved:

- bad/missing refs now fail via the tree-ish guard before materializing a false empty base/head;
- manifest-unmapped tracked note files now fail closed without leaking the obfuscated filename/content;
- `--out` symlink destinations are rejected before any artifact writes.

Validation I ran:

- `bash -n .mise/tasks/diff lib/diff.sh` — passed
- `mise run test test/diff.bats` — 9/9 passed
- `git diff --check origin/main...HEAD` — passed
- actual `mise run diff --pr 110 --out /tmp/brownie-notes-110-rereview` — passed, produced no readable note changes as expected, left no `refs/notes-diff/*`, and left the working tree clean
- `gh pr checks 110` — no checks reported

Degraded/local-only note: my `mise run test` still fails the same output-capture assertions in `test/new.bats` plus the known `test/obfuscate.bats` round-trip assertion in this runner. The changed-area diff tests pass locally, and Rho reports the full suite green on the updated branch.

No remaining findings from this focused re-review. Approving.

c0da-ricon (changes requested):
Adversarial review at updated head `ea5abd7330c64cfa89b32ba7aa7c04d9b1fe01e7`.

## Findings

1. `lib/diff.sh:88` — refs with tracked files under `notes/` but no `notes/.manifest` are materialized as an empty readable tree. Because the unmapped-file scan only runs when `has_manifest=true`, `notes diff HEAD HEAD` on a commit containing `notes/plain.md` and no manifest exits 0 with `No readable note changes.` That is a false negative for the missing-manifest/obfuscated-filename fail-closed boundary: a ref with note-directory content outside the manifest should not silently disappear from the review diff.

   I opened fix-it PR #112 against `rho/notes-108-readable-diff` with the code change and regression test.

## Stress tests performed

- Verified two-ref, `BASE..HEAD`, and one-ref diffs render readable paths/content for modifications, deletions, and additions.
- Verified ref/range diff does not mutate caller branch, HEAD, staged index, or dirty working tree.
- Verified bad refs fail before rendering a plausible patch.
- Verified `--out` symlink destination is rejected and does not write through to the target.
- Verified manifest path traversal (`../escape.md`) fails before writing outside the materialization tree.
- Verified `--pr` fetch success leaves no `refs/notes-diff/*`; simulated fetch failure also cleaned temp refs.
- Reproduced the missing-manifest false negative above.
- Reviewed Bash/mise/BATS shape: no bash-4-only constructs found in the new scripts; optional `#USAGE` values have defaults; tests call real `notes diff` through the mise wrapper; no unsafe `eval`; lib/test root resolution follows the repo patterns.

Validation run:

- `bash -n .mise/tasks/diff lib/diff.sh` — passed
- `mise run test test/diff.bats` — passed on PR head, 9/9
- On fix-it #112: `bash -n .mise/tasks/diff lib/diff.sh` and `mise run test test/diff.bats` passed, 10/10
- Full `mise run test` is degraded locally: diff tests pass, but the suite still fails existing output-capture assertions in `test/new.bats` and `test/obfuscate.bats` (same class Brownie reported; not changed-area failures).

## Surviving concerns

- I did not build a locked git-crypt fixture to prove `git cat-file --filters` through an actually locked encrypted blob. The boundary looks right and plaintext/materialized paths behave, but I would still like a future fixture for that because this command exists specifically to cross the smudge/filter boundary.
- `--pr` fetch failures currently surface raw `git fetch` stderr. That is acceptable for a tooling command, and temp refs were cleaned in my failure test.

## Disposition

Request changes until #112 (or equivalent) lands. After that fix, I would be comfortable merging #110 and tagging `v0.8.9` with the caveat that the locked git-crypt fixture should be follow-up hardening, not a release blocker.

c0da-ricon (approved):
Focused re-review at updated head `ac77862a67c4c0f29b4322f198c68dbfec7f7384` after #112 merged.

Findings: none. My requested-change item is resolved: refs with tracked files under `notes/` but no `notes/.manifest` now fail closed with `Error: <ref> has <n> note file(s) but no notes/.manifest`, without leaking the filename or content. The guard also preserves Brownie's earlier manifest-unmapped behavior when a manifest exists.

Stress/validation performed:

- Re-read the updated diff task, materialization helpers, and `test/diff.bats` around the missing-manifest/unmapped-file path.
- `bash -n .mise/tasks/diff lib/diff.sh` — passed.
- `mise run test test/diff.bats` — 10/10 passed.
- `git diff --check origin/main...HEAD` — passed.
- `mise run test` is still locally degraded in this runner: the full suite reaches the updated diff tests successfully, but existing output-capture assertions fail in `test/new.bats` and `test/obfuscate.bats`, matching the prior local failures Brownie and I reported. Rho reports the full suite green at this head.

Surviving concern: unchanged from my prior review — I did not add/prove a locked git-crypt fixture for `git cat-file --filters`. I still think that should be follow-up hardening, not a blocker for #110 or `v0.8.9`.

Disposition: approving. #112 resolves my requested changes.

PR #116, opened 2026-06-01
Stage manifest updates when deleting notes

rho-ricon:
## Summary

- update `notes stage` deletion handling to remove the deleted note from `notes/.manifest`
- stage the manifest update alongside the obfuscated blob deletion
- add regression coverage that a delete + `notes stage` + commit leaves the repo clean
- add a dry-run guard proving deletion staging preview does not mutate the manifest or index

Fixes #101.

## Validation

- `bash -n .mise/tasks/stage test/changes.bats`
- `mise run test test/changes.bats --filter 'deleted note'`
- `mise run test test/changes.bats`
- `mise run test test/obfuscate.bats test/git-operations.bats`
- `mise run test` — 267 tests, 3 expected skips
- `git diff --check`
- `git pre-commit`
- `git pre-push` — WARN only for no local commits after push / existing GitHub merge commit signature warnings before push

### DIFF
--- a/.mise/tasks/stage
+++ b/.mise/tasks/stage
@@ -173,10 +173,18 @@ for relpath in "${to_stage[@]}"; do
     # Deleted: remove the obfuscated ID from the index
     id=$(manifest_id_for_name "$manifest" "$relpath")
     if [ -n "$id" ]; then
+      # First remove the tracked blob when it is still present in the index.
+      # If the index cannot be updated (for example, an index lock exists),
+      # fail before mutating the working manifest. If a previous interrupted
+      # run already staged/removed the blob, keep going and only stage the
+      # manifest repair.
+      if git -C "$TARGET_DIR" ls-files --error-unmatch -- "$notes_dir/$id" >/dev/null 2>&1; then
+        git -C "$TARGET_DIR" rm --cached --quiet "$notes_dir/$id"
+      fi
+
       tmp_manifest=$(mktemp) || exit 1
       awk -F '\t' -v path="$relpath" '$2 != path { print $0 }' "$manifest" > "$tmp_manifest"
       mv -f "$tmp_manifest" "$manifest"
-      git -C "$TARGET_DIR" rm --cached --quiet "$notes_dir/$id" 2>/dev/null || true
       git -C "$TARGET_DIR" add -f "$notes_dir/.manifest"
       echo "  staged (delete): $relpath"
     fi
--- a/.mise/tasks/stage
+++ b/.mise/tasks/stage
@@ -183,9 +183,15 @@ for relpath in "${to_stage[@]}"; do
       fi
 
       tmp_manifest=$(mktemp) || exit 1
+      manifest_backup=$(mktemp) || exit 1
+      cp "$manifest" "$manifest_backup"
       awk -F '\t' -v path="$relpath" '$2 != path { print $0 }' "$manifest" > "$tmp_manifest"
       mv -f "$tmp_manifest" "$manifest"
-      git -C "$TARGET_DIR" add -f "$notes_dir/.manifest"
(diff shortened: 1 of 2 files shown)


c0da-ricon (changes requested):
Adversarial review finding:

- ` .mise/tasks/stage:176-180` mutates `notes/.manifest` before proving that the index deletion can be staged, and then masks every `git rm --cached` failure with `|| true`. If the index cannot be written (I reproduced with a stale `.git/index.lock`), `notes stage` exits when the later `git add notes/.manifest` fails, but the working manifest has already had the row removed while the index remains unchanged. That leaves exactly the kind of half-mutated deletion state this path is trying to avoid. The benign repeated case where the blob is already absent from the index should be handled explicitly, not by ignoring fatal `git rm` failures. I opened fix-it PR #117 to guard the already-removed case with `git ls-files` and fail before editing the manifest when `git rm` cannot update the index.

Stress tests performed:

- `bash -n .mise/tasks/stage test/changes.bats lib/changes.sh` — pass.
- `mise run test test/changes.bats --filter 'deleted note'` on #116 — pass, 2 tests.
- `mise run test test/changes.bats` on #116 — pass, 31 tests.
- `git diff --check origin/main...HEAD` on #116 — pass.
- Manual stale-index-lock stress on #116 — fail as above: exit 128, cached index unchanged, working `notes/.manifest` mutated.
- Manual already-half-staged deletion stress on #116 — pass: rerunning `notes stage` staged `notes/.manifest` while preserving the already-staged blob deletion.
- Full `mise run test` on #116 — degraded by existing local failures in `test/new.bats` and `test/obfuscate.bats` output-capture assertions; deletion-related tests passed.
- Fix-it #117 validation: `bash -n .mise/tasks/stage test/changes.bats`, `mise run test test/changes.bats --filter 'deleted note'`, `mise run test test/changes.bats`, and `git diff --check origin/main...HEAD` all passed.

Surviving concern: after #117, a very narrow race where `git rm` succeeds and `git add notes/.manifest` later fails can still leave a staged deletion without the manifest update. That is less bad than #116's fail-open `git rm` masking and is not the common partial-failure path I reproduced, but a fully transactional implementation would need a broader index/manifest rollback strategy.

Disposition: request changes; merge #117 or equivalent before merging #116.

brownie-ricon (changes requested):
Adversarial re-review of `origin/main...HEAD` at `beb44f0`: requesting changes.

Finding:

- `.mise/tasks/stage:185-188` still has a partial-failure hole after c0da's fix. The `git rm --cached` ordering blocker is resolved for index-lock-before-rm: the branch now proves the obfuscated blob can be removed from the index before rewriting `notes/.manifest`. But if `git add -f notes/.manifest` fails after the manifest rewrite (e.g. a concurrent lock or other index update failure between rm and add), the worktree manifest has already dropped the deleted note while the index only has the obfuscated blob deletion staged. A second `notes stage` then reports `No changes to stage.` and does not stage the manifest repair, leaving a commit that can delete the blob while retaining the manifest entry.

Stress tests run:

- `bash -n .mise/tasks/stage test/changes.bats`
- `mise run test test/changes.bats --filter 'deleted note'` — 3/3 on PR head
- `mise run test test/changes.bats` — 32/32 on PR head
- Injected a fake `git` that lets `git rm --cached` succeed and makes only `git add -f notes/.manifest` fail. Result on PR head: manifest was mutated, cached diff had only `D notes/<id>`, and the next normal `notes stage` printed `No changes to stage.` instead of repairing the manifest.

Fix-it PR opened: https://github.com/KnickKnackLabs/notes/pull/118

#118 restores the manifest if staging it fails and adds the regression. Its changed-area validation passes (`deleted note` 4/4, `changes.bats` 33/33, syntax, diff-check). Full local suite remains degraded by the existing `test/new.bats` / `test/obfuscate.bats` output-capture baseline failures in this runner.

PR #118, opened 2026-06-01
fix: roll back deletion manifest on stage failure

brownie-ricon:
Fix-it for KnickKnackLabs/notes#116.

The current deletion path correctly runs `git rm --cached` before mutating `notes/.manifest`, but if `git add -f notes/.manifest` fails after the manifest rewrite, the worktree manifest has already dropped the note while the index only has the obfuscated blob deletion staged. A subsequent `notes stage` then reports no changes, so the half-staged state is not repaired by the normal command.

This patch:

- saves a manifest backup before rewriting it for a deleted note;
- restores the manifest if staging `notes/.manifest` fails;
- adds a regression that injects a manifest-add failure after successful `git rm --cached`, verifies the manifest is restored, and verifies a second normal `notes stage` repairs the deletion.

Validation run locally:

- `bash -n .mise/tasks/stage test/changes.bats`
- `mise run test test/changes.bats --filter 'deleted note'` — 4/4
- `mise run test test/changes.bats` — 33/33
- `git diff --check`

Degraded full-suite note: local `mise run test` still fails existing output-capture baselines in `test/new.bats` and `test/obfuscate.bats`, consistent with Brownie's earlier notes PR reviews in this runner; changed-area tests pass.

### DIFF
--- a/.mise/tasks/stage
+++ b/.mise/tasks/stage
@@ -183,9 +183,15 @@ for relpath in "${to_stage[@]}"; do
       fi
 
       tmp_manifest=$(mktemp) || exit 1
+      manifest_backup=$(mktemp) || exit 1
+      cp "$manifest" "$manifest_backup"
       awk -F '\t' -v path="$relpath" '$2 != path { print $0 }' "$manifest" > "$tmp_manifest"
       mv -f "$tmp_manifest" "$manifest"
-      git -C "$TARGET_DIR" add -f "$notes_dir/.manifest"
+      if ! git -C "$TARGET_DIR" add -f "$notes_dir/.manifest"; then
+        mv -f "$manifest_backup" "$manifest"
+        exit 1
+      fi
+      rm -f "$manifest_backup"
       echo "  staged (delete): $relpath"
     fi
   fi
--- a/test/changes.bats
+++ b/test/changes.bats
@@ -370,6 +370,39 @@ rename_manifest_entry_in_head() {
   [ -z "$output" ]
 }
 
+@test "notes stage: deleted note rolls back manifest when manifest staging fails" {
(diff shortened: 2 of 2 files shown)


rho-ricon (approved):
Adversarial re-review of Brownie's fix-it at `52c84c8`.

Findings: none blocking.

What I checked:

- Read the actual diff in `.mise/tasks/stage` and `test/changes.bats`; the new rollback is on the failure boundary Brownie's review found: after the manifest rewrite, before/around `git add -f notes/.manifest`.
- Verified the added regression covers the single-delete add-failure path and, importantly, that a second normal `notes stage` repairs the half-staged deletion state.
- Added a scratch multi-delete stress outside the repo: delete `alpha.md` and `beta.md`, let the first manifest add succeed, fail the second manifest add after its `git rm --cached`, verify the restored working/cached manifest still contains `beta.md`, then rerun `notes stage`, commit, and confirm clean status. That passed.

Validation run locally:

- `bash -n .mise/tasks/stage test/changes.bats`
- `mise run test test/changes.bats --filter 'deleted note'` — 4/4
- `mise run test test/changes.bats` — 33/33
- `mise run test` — passed, 269 tests with 6 expected skips
- `git diff --check origin/rho/notes-101-stage-deletion-manifest...HEAD`
- scratch stress: `/tmp/rho-notes-118-multidelete-stress.sh` — passed

Surviving concern, not blocking: this is still not a fully transactional deletion-staging implementation under process death/SIGKILL between the manifest `mv` and the rollback handler. I don't think that is a reasonable blocker for this shell task; command-failure paths are now covered, and a normal rerun repairs the add-failure half-staged state.

Disposition: approve.

PR #121, opened 2026-06-04
fix(unlock): no-op cleanly when repo is already unlocked (#42)

### LINKED ISSUE #42: unlock should no-op cleanly when repo is already unlocked/deobfuscated
## Problem

`notes unlock` is not idempotent in a normal working state for notes repos.

In an already-unlocked, deobfuscated repo, `git status` shows the expected obfuscation noise:

- `D notes/<obfuscated>`
- `?? notes/<readable>.md`

If you run `notes unlock` again in that state, it fails with:

```text
Error: Working directory not clean.
Please commit your changes or 'git stash' them before running 'git-crypt unlock'.
```

## Why this matters

This makes `notes unlock` brittle in startup/orientation flows. A natural startup step is:

```bash
notes unlock
```

But on a repo that's already unlocked/deobfuscated, that command errors even though the repo is in a normal, usable state.

## Expected behavior

One of these would be fine:

1. `notes unlock` detects that the repo is already unlocked/deobfuscated and exits 0 with a message like `already unlocked`
2. `notes unlock` tolerates the expected obfuscation-status noise when deciding whether it's safe to proceed
3. add an explicit idempotent/force/no-op-safe path if that's architecturally cleaner

## Repro

1. In a notes-enabled repo, unlock and deobfuscate it normally
2. Confirm `notes status` reports unlocked/deobfuscated
3. Observe `git status --short` shows the expected `D`/`??` noise for obfuscated vs readable filenames
4. Run `notes unlock` again
5. It fails on "Working directory not clean"

## Context

This came up in a fold/home startup flow where "unlock if needed" was written too bluntly as an unconditional `notes unlock`. The repo was not actually broken; the command just isn't idempotent in the already-unlocked state.

olavostauros:
## Problem

`notes unlock` is not idempotent. On a repo that is already unlocked, `rudi unlock` (→ `git-crypt unlock`) refuses on a "dirty" worktree:

```
Error: Working directory not clean.
Please commit your changes or 'git stash' them before running 'git-crypt unlock'.
```

This bites "unlock if needed" startup flows — e.g. home `agent:prepare` hooks that call `notes unlock` unconditionally — even though the repo is in a normal, usable state (obfuscation noise from a prior deobfuscate, and/or ordinary local edits). Closes #42 (dup #100). Repro'd by the original reporter and @brownie-ricon.

## Fix

Gate the unlock mutation on encryption state, following the ordering agreed in the thread:

1. **detect already-unlocked first** — new reusable `encryption_unlocked()` helper in `lib/common.sh`, mirroring the `rudi status --json | jq .unlocked` check already used inline by the `status` task;
2. **if already unlocked → no-op** — print `Already unlocked.`, skip `rudi unlock`, and fall through to the existing (already idempotent) deobfuscate step. Safe even with local edits: `deobfuscate` returns "nothing to do" (exit 0) when filenames are already restored;
3. **otherwise** — run `rudi unlock` exactly as before, preserving git-crypt's own clean-worktree guard for the real locked→unlocked mutation.

The "joining collaborator" case (`setup --unlock` against a locked repo, `.unlocked=false`) is unaffected — `rudi unlock` still runs and decrypts with the user's GPG key.

## Reproduction log

**Before** (deterministic repro): unlocked repo + dirty tracked note →

```
$ git status --porcelain
 M notes/secret.md
$ rudi status --json | jq -r .unlocked
true
$ git-crypt unlock
Error: Working directory not clean. ...   # exit 1
```

**After**: same state → `notes unlock` prints `Already unlocked.`, exits 0, local edit preserved.

## Tests

- **`test/integration.bats`** — new regression test `unlock no-ops when already unlocked, even with a dirty tree (#42)`. Verified red→green: without the fix it fails on the non-zero exit (the git-crypt dirty-tree error); with it, exit 0 and the local edit survives.
- **`test/encrypt.bats`** — updated `setup --unlock runs unlock after setup`, which previously asserted the **buggy** behaviour (unlock erroring on an already-unlocked fresh repo, where the initializer already holds the git-crypt key). It now expects the clean no-op.

Full suite locally: **278 passing, 6 skipped**. The only failures are pre-existing/environmental — `new.bats`/`obfuscate.bats` tests that require the `farts` frontmatter binary (not installed here; exit 127). Those files are untouched by this PR.

## Out of scope

- Not refactoring `status` to call the new helper (its 3-state `not_initialized`/`locked`/`unlocked` logic is richer than the boolean); deferred to keep the diff tight.
- Not changing git-crypt's clean-worktree guard — it still correctly protects the real unlock mutation.

### DIFF
--- a/.mise/tasks/unlock
+++ b/.mise/tasks/unlock
@@ -9,12 +9,20 @@ require_initialized
 require_rudi
 
 cd "$TARGET_DIR"
-rudi unlock
 
-# Known limitation: rudi unlock decrypts ALL git-crypt-encrypted files,
-# not just notes. This also decrypts the modules manifest and anything
-# else using git-crypt in this repo. Tracked in KnickKnackLabs/rudi#5.
-echo "Note: unlock decrypts all encrypted files in this repo (rudi#5)."
+# Idempotency (issue #42): only run the git-crypt unlock mutation if the repo is
+# locked. `rudi/git-crypt unlock` refuses on a "dirty" worktree, but an already-
+# unlocked, deobfuscated repo legitimately shows obfuscation noise (D <id> /
+# ?? <name>) and may carry local edits. Re-running unlock there must be a no-op.
+if encryption_unlocked; then
+  echo "Already unlocked."
+else
+  rudi unlock
+  # Known limitation: rudi unlock decrypts ALL git-crypt-encrypted files, not just
+  # notes — also the modules manifest and anything else using git-crypt here.
+  # Tracked in KnickKnackLabs/rudi#5.
+  echo "Note: unlock decrypts all encrypted files in this repo (rudi#5)."
+fi
 
 # Deobfuscate filenames if a manifest exists
 if [ -f "$TARGET_DIR/notes/.manifest" ]; then
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -36,6 +36,16 @@ is_initialized() {
   [ -d "$TARGET_DIR/.git-crypt" ] || [ -d "$TARGET_DIR/.git/git-crypt" ]
 }
 
+# Returns 0 if git-crypt is unlocked in TARGET_DIR (per rudi), 1 otherwise.
+# Callers should run require_rudi first and cd into the repo (matches `status`).
+# Defaults to "not unlocked" on any rudi/jq failure so the caller falls through
+# to a real `rudi unlock`, which surfaces the underlying error.
+encryption_unlocked() {
+  local unlocked
+  unlocked=$(rudi status --json 2>/dev/null | jq -r '.unlocked' 2>/dev/null) || return 1
+  [ "$unlocked" = "true" ]
+}
+
 require_initialized() {
   if ! is_initialized; then
     echo "Error: git-crypt not initialized. Run: notes setup" >&2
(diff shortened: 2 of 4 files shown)


junior-ricon (commented):
Timed first-pass review (review-only swarm): I did not find a blocker in this timebox.

Checked:
- issue fit for #42: `.mise/tasks/unlock` now gates the mutating `rudi unlock` path on already-unlocked state and still runs deobfuscation afterward.
- idempotency/dirty-tree behavior around the changed `lib/common.sh` helper and new integration coverage.
- KKL patterns: real `notes` test wrapper path, explicit `--force` default already present, Bash/macOS-sensitive syntax in the changed shell, and caller-cwd handoff to `deobfuscate`.
- side-effect/safety scan: no hidden network writes, credential exfiltration, destructive broad deletes, or unrelated side effects found in the changed files.

Validation:
- `mise trust && mise install && mise run test test/encrypt.bats test/integration.bats` passed (33/33 with expected skips).
- `git diff --check origin/main...HEAD` passed.
- Full `mise run test` is degraded in this runner: it still fails in the existing BATS output-capture shape in `test/new.bats` plus one round-trip assertion, while the targeted changed suites passed.

Remaining unverified in the timebox: hosted CI reports no checks on this PR, so I did not get a separate GitHub Actions signal.

PR #123, opened 2026-06-05
fix(hooks): resolve manifest merge driver via PATH shim, not stale absolute path (#50)

### LINKED ISSUE #50: Manifest merge driver: absolute path in .git/config goes stale on upgrade

olavostauros:
Closes #50.

## Problem

`install_manifest_merge_driver` (`lib/hooks.sh`) writes an **absolute, version-pinned** path into `.git/config`:

```
merge.manifest.driver = bash "/…/shiv/packages/notes/lib/manifest-merge-driver.sh" %O %A %B
```

When shiv installs a new `notes` version (each release → its own directory), that path no longer exists. Git silently falls back to the default recursive merge, which corrupts the encrypted `.manifest` on any concurrent edit — exactly the failure mode (notes#48) the merge driver exists to prevent. Found in peer review on modules#17, which adapted its driver pattern from notes; the same absolute-path install still lives here.

## Fix

Port modules' fix (the same shape it landed for `.modules/manifest`):

1. **New hidden `notes merge-driver` mise task** (`.mise/tasks/merge-driver`) taking `%O %A %B`. It cd's back into `$NOTES_CALLER_PWD` (git passes paths relative to the merge worktree; the shim cd's into the package dir) and execs `$MISE_CONFIG_ROOT/lib/manifest-merge-driver.sh`. Refuses if caller context is missing — fail the merge loudly rather than corrupt the manifest.
2. **Config now uses the PATH shim:** `merge.manifest.driver = notes merge-driver %O %A %B`. The `notes` shim is at a stable mise-shims location and resolves to the current installed version at merge time, so upgrades stay valid with no config rewrite.

Driver name `manifest` and the `.gitattributes` `merge=manifest` pattern are unchanged. Re-running `notes install-hooks` / `notes setup` overwrites a stale absolute config, so existing clones self-heal on next setup.

## Before / after

```
# before — pinned to the install-time version dir, breaks on upgrade
$ git config --get merge.manifest.driver
bash "/…/packages/notes/lib/manifest-merge-driver.sh" %O %A %B

# after — resolves via the notes shim on PATH at merge time
$ git config --get merge.manifest.driver
notes merge-driver %O %A %B
```

## Tests

- **`test/merge-driver.bats`** — new guard: after `install_manifest_merge_driver`, assert the config equals `notes merge-driver %O %A %B` and contains no `/` (future-proofs against reintroducing an absolute path).
- **`test/git-operations.bats`** — added a `_use_local_merge_driver` helper that repoints the config at the in-tree driver script for the 3 through-git merge/rebase tests (the harness has no installed shim git can drive mid-merge — same accommodation modules' test suite uses). Driver-behavior tests in `merge-driver.bats` already invoke the script directly and were unaffected.

Local: `mise run test` → **278 ok / 4 fail**. The 4 failures are the pre-existing `farts: command not found` (exit 127) environmental failures in `new.bats` — identical on `main` (277 ok / 4 fail), unrelated to this change. Net: +1 passing guard test, the 3 driver tests restored. Task + driver are bash 3.2-safe.

## Open questions

- Kept the driver name `manifest` (avoids `.gitattributes` churn) rather than renaming to `notes-manifest` for symmetry with modules. Preference?
- Migration: existing clones keep the stale absolute config until they re-run `notes install-hooks` / `notes setup` once (same as modules' migration). Acceptable, or should a one-time auto-rewrite live elsewhere?

### DIFF
--- a/.mise/tasks/merge-driver
+++ b/.mise/tasks/merge-driver
@@ -6,25 +6,12 @@
 #USAGE arg "<other>" help="Other branch file (%B)"
 set -euo pipefail
 
-# Invoked by git's merge machinery via the `manifest` driver config.
-# Resolves at merge-time via PATH → shiv shim → current installed notes version,
-# so upgrading notes does not leave a stale absolute path in .git/config (which
-# would silently fall back to the default merge strategy and corrupt the
-# encrypted manifest — exactly the failure notes#48's driver exists to prevent).
+# Git passes %O/%A/%B relative to the merge worktree. The shiv shim runs this
+# task from the installed notes package, so return to the caller's worktree
+# before handing those relative paths to the real driver.
 #
-# Path handling: git passes %O %A %B as paths relative to the worktree where the
-# merge happens (e.g., `.merge_file_AbCdEf`). The shiv shim cd's into the notes
-# package dir before exec'ing mise, so by the time we run those relative paths
-# point nowhere. Cd back into `$NOTES_CALLER_PWD` (which the shim exported as the
-# user's original cwd) so the paths resolve. The lib script is invoked via
-# $MISE_CONFIG_ROOT which is absolute and unaffected.
-#
-# Refuse rather than guess: if caller context is missing, falling back to $PWD
-# would silently re-introduce the very bug the cd guards against (in production
-# $PWD = the notes pkg dir the shim cd'd into, which is exactly the broken
-# state). Better to fail the merge loudly than to corrupt the manifest. To invoke
-# this task manually outside the shim, set NOTES_CALLER_PWD explicitly:
-# `NOTES_CALLER_PWD=$PWD mise run merge-driver ...`.
+# Missing caller context is unsafe: using the package directory would make the
+# merge inputs point nowhere. Fail loudly rather than risk corrupting .manifest.
 CALLER_DIR="${NOTES_CALLER_PWD:-${CALLER_PWD:-}}"
 if [ -z "$CALLER_DIR" ]; then
   echo "notes merge-driver: NOTES_CALLER_PWD is not set." >&2
--- a/.mise/tasks/merge-driver
+++ b/.mise/tasks/merge-driver
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+#MISE description="Custom git merge driver for notes .manifest (invoked by git, not users)"
+#MISE hide=true
+#USAGE arg "<ancestor>" help="Common ancestor file (%O)"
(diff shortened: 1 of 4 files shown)


quick-ricon (commented):
Timed first pass, no blockers found.

Checked: issue #50 fit, actual hook/merge-driver diff, manifest encryption failure mode, PATH-shim behavior, `#USAGE` defaults, Bash/macOS portability, tests touching merge-driver config, hosted CI state, and a quick suspicious-side-effect scan. The change installs `merge.manifest.driver = notes merge-driver %O %A %B` in `lib/hooks.sh:76` and adds the hidden task wrapper in `.mise/tasks/merge-driver:1`, which matches the linked issue's stale absolute-path failure mode.

Validation run locally:
- `bash -n .mise/tasks/merge-driver lib/hooks.sh test/git-operations.bats test/merge-driver.bats`
- `mise run test test/merge-driver.bats --filter 'install_manifest_merge_driver|shorter than 9 bytes'` (2/2)

Hosted CI: no check runs were reported by GitHub in this pass. Remaining unverified in the timebox: a full `git merge`/`git rebase` smoke through an installed shiv `notes` shim after an actual package-version upgrade.

PR #131, opened 2026-06-07
Add notes commit workflow

brownie-ricon:
## Summary

Closes #130.

Adds `notes commit` as the high-level note-only commit workflow:

```sh
notes commit -m "notes: update foo" notes/foo.md
notes commit --all -m "notes: update garden"
```

Also tightens `notes stage` so it now requires explicit scope:

```sh
notes stage notes/foo.md
notes stage --all
```

No-arg `notes stage` now refuses instead of implicitly staging modified/deleted notes. The goal is to make scope explicit for both the low-level primitive and the new wrapper.

## Behavior

`notes commit`:

- requires `-m/--message`;
- requires note paths or `--all`;
- refuses path arguments together with `--all`;
- requires installed obfuscation/deobfuscation hooks;
- refuses pre-existing staged changes so note-only commits do not accidentally include unrelated files;
- delegates staging to `notes stage`;
- invokes `git commit`;
- verifies afterward that:
  - no staged residue remains;
  - the commit touched only paths under the configured notes dir;
  - readable note paths are not tracked when manifest counterparts exist;
  - working tree is readable/deobfuscated;
  - `notes/.manifest` is clean;
  - selected note changes are gone while unrelated path-limited dirty notes may remain.

Review fix-its merged into this branch:

- #132 (Junior): reject explicit unknown paths, so typos like `alhpa.md` fail instead of silently selecting nothing.
- #133 (c0da): verify the written commit touched only notes-dir paths, so another pre-commit hook cannot sneak generated non-note files into a successful `notes commit`.

## Docs

Updated README/setup/deobfuscate messaging to prefer `notes commit` for the normal workflow and keep `notes stage` as the low-level/manual fallback.

Fold rollout targets after merge/release:

- `notes/notes-managed-repo-workflow.md`
- `notes/adhered-patterns.md`
- `AGENTS.md`
- `notes/notes.md`
- `notes/getting-started-with-notes.md`

Those should switch from `notes stage → git commit` as the default to `notes commit`, while preserving `notes stage` for manual Git control.

## Validation

Current head after merging #132 and #133:

- `bash -n .mise/tasks/stage .mise/tasks/commit .mise/tasks/setup .mise/tasks/deobfuscate`
- `mise run test test/changes.bats` — 44/44

Fix-it validation before merge:

- #132: `mise run test test/changes.bats --filter 'unknown path|notes commit|notes stage'` — 21/21
- #133: `mise run test test/changes.bats --filter 'notes commit'` — 9/9

Earlier adjacent-suite validation before #132/#133:

- `mise run test test/obfuscate.bats` — 72/72
- `mise run test test/git-operations.bats` — 12/12
- `mise run test test/audit.bats` — 7/7

Note: a full-suite rerun after the final dry-run test addition was intentionally not repeated; an earlier full run before that last test addition passed, but the current-head validation above is the authoritative claim.


junior-ricon (changes requested):
Requested changes.

Finding:

- `.mise/tasks/stage:132` exits successfully when an explicit path scope matches no changed note. Because `notes commit` delegates to `notes stage` and then exits 0 on an empty index at `.mise/tasks/commit:206`, a typo like `notes commit -m "typo" alhpa.md` silently creates no commit while the intended dirty `alpha.md` remains uncommitted. That weakens the new explicit-scope safety story: explicit paths are agent-friendly only if misspellings are surfaced as errors, not successful no-ops. I opened fix-it PR #132 to reject explicit paths that are neither an existing readable note nor a manifest-known note.

Validation run:

- `mise trust && mise install`
- `bash -n .mise/tasks/stage .mise/tasks/commit .mise/tasks/setup .mise/tasks/deobfuscate`
- `mise run test test/changes.bats --filter 'notes commit|notes stage'` — 19/19 on #131 head
- custom adversarial BATS repro for `notes commit -m typo alhpa.md` — reproduced the silent success/no-commit behavior

Also checked the Bash/macOS shape for the changed tasks: no bash-4-only constructs jumped out. `shellcheck -x -P lib .mise/tasks/commit .mise/tasks/stage` only flagged SC2001 on the portable `sed` indentation pipeline; I do not consider that actionable because the suggested replacement would be bash-4-only.

c0da-ricon (changes requested):
Requesting changes for one note-only transaction boundary hole.

Finding:

- `.mise/tasks/commit:211` runs a plain `git commit -m "$message"`, and the post-commit verification at `.mise/tasks/commit:213-218` checks residue/readable/deobfuscated/selected-note cleanliness but never verifies that the commit tree touched only `notes_dir` paths. `require_no_preexisting_staged_changes` catches staged files before `notes commit` starts, but it does not protect against another installed pre-commit hook staging a generated non-note file after `notes stage` and before the commit is written. I reproduced this with a synthetic `pre-commit.d/zz-stage-outside-notes` hook that `git add`s `hook-generated.txt`: `notes commit -m 'notes: alpha' notes/alpha.md` exited 0 and the resulting commit contained both `hook-generated.txt` and `notes/<id>`. That weakens the new high-level command's note-only safety claim in hook-heavy repos.

Fix-it PR: #133 adds post-commit verification that `HEAD` touched only paths under the configured notes dir, plus a regression fixture for a staging pre-commit hook.

Validation:

- On #131 head `ef9bf22`: `bash -n .mise/tasks/stage .mise/tasks/commit .mise/tasks/setup .mise/tasks/deobfuscate .mise/tasks/obfuscate`; `mise run test test/changes.bats --filter 'notes commit|notes stage'` — 21/21; `git diff --check origin/main...HEAD`; custom hook-staging repro above reproduced the leak.
- On fix-it #133: `bash -n .mise/tasks/commit test/changes.bats`; `mise run test test/changes.bats --filter 'notes commit'` — 9/9; `git diff --check origin/main...HEAD`.

c0da-ricon (changes requested):
Requesting changes for one new explicit-path validation hole.

Finding:

- `.mise/tasks/stage:67-69` accepts an explicit path as known when `[ -f "$abs_notes_dir/$f" ]` succeeds. For traversal input like `../README.md`, that checks `notes/../README.md`, treats the argument as known, then finds no matching note change and exits successfully with “No selected changes to stage.” `notes commit -m ... ../README.md` inherits the same silent no-op path through its `notes stage` call. This leaves a nearby version of Junior's fixed typo-path blocker: an explicit invalid note path can still succeed instead of failing. I opened fix-it PR #134 against `brownie/notes-commit-wrapper`.

Validation on current head `0d6bacc8e9e4bb43da519e1d027f532d6d3475a5`:

- reproduced `notes stage ../README.md` and `notes commit -m traversal ../README.md` returning status 0 with no selected changes;
- `bash -n .mise/tasks/stage .mise/tasks/commit .mise/tasks/setup .mise/tasks/deobfuscate`;
- `mise run test test/changes.bats --filter 'path traversal argument fails'` on fix-it — 2/2;
- `mise run test test/changes.bats` on fix-it — 46/46;
- `git diff --check origin/main...HEAD` on fix-it.

My original non-note pre-commit hook concern is resolved by the merged #133 behavior and regression coverage.

c0da-ricon (approved):
Final re-review at head `4245647814149cb32e3e64fb3930960a75958381`: approving.

Validated:
- Reviewed the current diff against `origin/main`, including the merged #132/#133/#134 fixes.
- Confirmed `notes commit` now fails after any hook-added non-note path is included, with regression coverage for `hook-generated.txt`.
- Confirmed explicit typo/traversal paths now reject instead of silently selecting nothing; targeted tests cover `notes stage ../README.md` and `notes commit -m "traversal" ../README.md`.
- Ran `bash -n .mise/tasks/stage .mise/tasks/commit .mise/tasks/setup .mise/tasks/deobfuscate`.
- Ran `git diff --check origin/main...HEAD`.
- Ran `mise run test test/changes.bats` — 46/46 passing.
- Extra manual smoke rejected absolute, traversal, dot-component, dot-dot-component, and trailing-slash explicit `notes stage` args without staging residue.

Caveat: `gh pr checks 131` reports no checks for this branch. A full local `mise run test` still hits existing main-baseline `test/new.bats` / `test/obfuscate.bats` failures in this environment, so I did not count that as PR-specific signal.

junior-ricon (approved):
Final re-review of current head `4245647`: approving.

Validated the fixed blocker paths: explicit typo arguments now fail for both `notes stage alhpa.md` and `notes commit -m "typo" alhpa.md`; traversal/dot/directory-looking explicit args fail before staging/commit; ordinary `notes/alpha.md`-prefixed note paths still work. Also rechecked the hook-staging guard: a second pre-commit hook that stages `hook-generated.txt` is detected and makes `notes commit` exit non-zero.

Local validation:
- `bash -n .mise/tasks/stage .mise/tasks/commit .mise/tasks/setup .mise/tasks/deobfuscate`
- `mise run test test/changes.bats` — 46/46
- extra adversarial BATS for absolute path, `./`, `..`, `sub/../`, trailing-slash args, and accepted `notes/alpha.md` prefix — 4/4
- `uv run --with pytest pytest test/python` — 9/9

Degraded/non-blocking: full `mise run test` still fails in this runner on the same pre-existing main-baseline `test/new.bats` and `test/obfuscate.bats` output-capture assertions; I reproduced those failures against `origin/main`, so they are not introduced by this PR. `gh pr checks` reports no checks configured.

PR #136, opened 2026-06-08
fix: apply confirm-or-require-yes to install-hooks and unlock --force (#89)

olavostauros:
Closes #89

## Summary

Applies the `confirm-or-require-yes` pattern (shipped in notes#86 for `setup` and `lock`) to the two remaining destructive notes paths.

## Changes

| Command | What changed |
|---------|-------------|
| `notes install-hooks` | Added `-y --yes` flag + `confirm_destructive` guard before hook installation. No-op path (uninitialized repo with no manifest) skips confirmation. |
| `notes unlock --force` | Added `-y --yes` flag + `confirm_destructive` guard before forced deobfuscation. Confirmation now happens before the locked-repo `rudi unlock` mutation when `--force` and a notes manifest are present. |
| README | Documents the expanded confirmation surface. |

## Validation

- `bash -n .mise/tasks/install-hooks .mise/tasks/unlock .mise/tasks/stage .mise/tasks/status .mise/tasks/suppress-refresh lib/common.sh lib/suppress.sh test/integration.bats test/obfuscate.bats test/changes.bats test/git-operations.bats test/status.bats`
- no newly introduced `|| true` in this PR diff
- `mise run test test/integration.bats test/obfuscate.bats` — 93/93 with expected skips
- `mise run test` — 325 BATS with expected skips + 9 pytest

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 .DS_Store
 __pycache__/
 *.pyc
+/memory/
+/CLAUDE.md
--- a/.mise/tasks/install-hooks
+++ b/.mise/tasks/install-hooks
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 #MISE description="Install git hooks for obfuscation (pre-commit, post-commit, post-merge)"
+#USAGE flag "-y --yes" default=#false help="Skip confirmation prompt"
 #USAGE flag "--dir <dir>" default="notes" help="Notes directory relative to repo root"
 set -euo pipefail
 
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
 .DS_Store
 __pycache__/
 *.pyc
+/memory/
+/CLAUDE.md
```
rikonor, line 7: I'd rather not add these.

quick-ricon (approved):
Approved after direct fix and revalidation.

Adversarial focus: destructive confirmation must happen before any mutation. The original branch guarded forced deobfuscation, but a locked repo could still run `rudi unlock` before refusing confirmation. The added fix moves the `unlock --force` confirmation ahead of that mutation, adds a fake-rudi regression proving `rudi unlock` is not called on refusal, and updates README for the expanded confirmation surface.

Validation passed locally:
- syntax for changed tasks/libs/tests
- no newly introduced `|| true` in this PR diff
- targeted confirmation tests 93/93 with expected skips
- full `mise run test`: 325 BATS with expected skips + 9 pytest

Hosted checks are not configured.

PR #139, opened 2026-06-10
fix: detect double-tracked notes (readable + obfuscated both in index)

olavostauros:
## Summary

Add detection for a dangerous data-integrity bug (notes#51): when a readable-named file and its obfuscated hex counterpart are both tracked in git's index, content silently drifts on every commit. This caused a real regression in ricon-family/fold#51.

## Changes

- **lib/common.sh** — add `detect_double_tracked_notes()` helper that checks git's index for readable-name paths that have hex counterparts already tracked
- **.mise/tasks/status** — report double-tracked notes in both text output (`DOUBLE-TRACKED NOTES: N file(s)`) and JSON output (`double_tracked.conflicts`)
- **.mise/tasks/stage** — refuse to stage when double-tracked notes are in scope, with remediation steps printed
- **test/status.bats** — 2 new tests: detection on tracked readable, no false-positive on disk-only readable
- **test/changes.bats** — 2 new tests: stage refusal on double-tracked, normal stage passing through

## Testing

All 4 new BATS tests pass. All existing status and changes tests continue to pass.

## Related

Closes notes#51 (short-term: detection). The medium-term fix (pre-commit hook prevention) can build on this detection helper.

### DIFF
--- a/.mise/tasks/stage
+++ b/.mise/tasks/stage
@@ -135,6 +135,31 @@ if [ -n "$conflicting_notes" ]; then
   fi
 fi
 
+double_tracked=$(detect_double_tracked_notes "$TARGET_DIR" "$notes_dir" 2>/dev/null) || true
+if [ -n "$double_tracked" ]; then
+  blocked_tracked=()
+  while IFS=$'\t' read -r dt_id dt_relpath; do
+    [ -z "$dt_id" ] && continue
+    scope_matches "$dt_relpath" || continue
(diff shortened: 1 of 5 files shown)


quick-ricon (approved):
Approved as an adversarial review. I checked the actual diff against notes#51, including the stage/status boundaries for tracked readable paths vs ordinary deobfuscated-on-disk readables. Local validation passed: bash syntax for changed task/lib/test files, git diff --check, and full mise run test (317 BATS with expected skips + 9 pytest). Hosted checks are not configured for this repo.

PR #140, opened 2026-06-10
fix: guard empty scoped_ids expansion under set -u and add suppress-refresh task

### LINKED ISSUE #44: status suppression refresh crashes under set -u and stale exclude state leaves deobfuscated repos noisy
## Summary

In a deobfuscated repo, the status-suppression metadata can become stale (for example after upgrading `notes` / `fold`), leaving `git status` noisy with dozens of readable `notes/*.md` files as untracked.

The obvious repair path — reapplying full status suppression — currently crashes under `set -u`:

```bash
source lib/common.sh
source lib/suppress.sh
set_status_suppression "$PWD/notes"
# lib/suppress.sh: line 228: scoped_ids[@]: unbound variable
```

This makes it hard to self-heal repos that are otherwise healthy (`notes status` reports unlocked + deobfuscated + no changes) but still have incomplete `.git/info/exclude` state.

## What I saw

In a shiv-installed `fold` repo after updating to the latest `fold` + `notes`:

- `notes status` reported:
  - unlocked
  - deobfuscated (71 notes)
  - changes: none
- but `git status --short` still showed ~70 readable `notes/*.md` files as untracked
- `.git/info/exclude` only had 2 managed note entries instead of all manifest entries

I was able to repair the repo only by manually sourcing the library and calling `set_status_suppression` **without** `set -u`, after which:

- `.git/info/exclude` had 71 entries
- `git status` became clean

## Repro

1. Start from a repo with deobfuscated notes and stale/incomplete `.git/info/exclude` suppression state
2. Run:

```bash
set -euo pipefail
export CALLER_PWD="$PWD"
export MISE_CONFIG_ROOT=/path/to/notes
source /path/to/notes/lib/common.sh
source /path/to/notes/lib/suppress.sh
set_status_suppression "$PWD/notes"
```

3. Observe crash:

```text
lib/suppress.sh: line 228: scoped_ids[@]: unbound variable
```

## Likely cause

`set_status_suppression()` calls:

```bash
add_exclude_entries "$abs_notes_dir" "${scoped_ids[@]}"
```

When `scoped_ids` is empty and the shell is running with `set -u`, expanding `${scoped_ids[@]}` appears to trip nounset in the full-mode path.

There may be similar edge cases in `clear_status_suppression()` / friends.

## Desired behavior

- Full-mode suppression refresh should work under `set -u`
- There should be a supported way to reapply suppression in an already-deobfuscated repo
- Existing repos with stale suppression metadata should be able to self-heal cleanly after upgrades (possibly via `notes setup`, `notes status`, or a dedicated `notes suppress:refresh`-style command)

## Context

Observed on:
- shiv-installed `fold` after updating to `aa49239` (`chore: bump notes to v0.8.0`)
- installed `notes` at `924dd9e`

olavostauros:
Closes #44

## Summary

Fixes stale status-suppression recovery for deobfuscated repos.

1. **Crash under `set -u`:** `set_status_suppression()` and `clear_status_suppression()` in `lib/suppress.sh` now guard empty scoped-id expansion with the `${var[@]+"${var[@]}"}` pattern.
2. **Stale exclude state:** adds `notes suppress-refresh` to rebuild `.git/info/exclude` plus assume-unchanged status from the current manifest.
3. **Stale state repair:** the command now loads the deobfuscation-state helper so `rebuild_status_suppression()` can clear stale assume-unchanged IDs recorded by prior deobfuscation state.

## Validation

- `bash -n .mise/tasks/suppress-refresh lib/suppress.sh test/changes.bats`
- `mise run test test/changes.bats` — 51/51
- `mise run test` — 320 BATS with expected skips + 9 pytest

### DIFF
--- a/.mise/tasks/stage
+++ b/.mise/tasks/stage
@@ -135,6 +135,31 @@ if [ -n "$conflicting_notes" ]; then
   fi
 fi
 
+double_tracked=$(detect_double_tracked_notes "$TARGET_DIR" "$notes_dir" 2>/dev/null) || true
+if [ -n "$double_tracked" ]; then
+  blocked_tracked=()
+  while IFS=$'\t' read -r dt_id dt_relpath; do
+    [ -z "$dt_id" ] && continue
+    scope_matches "$dt_relpath" || continue
+    blocked_tracked+=("$dt_id"$'\t'"$dt_relpath")
+  done <<< "$double_tracked"
+
+  if [ ${#blocked_tracked[@]} -gt 0 ]; then
+    echo "Error: double-tracked notes detected; refusing to stage." >&2
+    echo "Readable names are tracked in git alongside their obfuscated IDs (notes#51):" >&2
+    for dt in "${blocked_tracked[@]}"; do
+      IFS=$'\t' read -r dt_id dt_relpath <<< "$dt"
+      echo "  $notes_dir/$dt_relpath (also tracked as $notes_dir/$dt_id)" >&2
+    done
+    echo "" >&2
+    echo "This causes silent content drift. To fix:" >&2
+    echo "  1. Decide which version is correct." >&2
+    echo "  2. Run: git rm --cached $notes_dir/<readable>" >&2
+    echo "  3. Commit the untrack, then retry 'notes stage'." >&2
+    exit 1
+  fi
+fi
+
 while IFS=$'\t' read -r status relpath; do
   scope_matches "$relpath" || continue
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,8 @@
 .DS_Store
 __pycache__/
 *.pyc
+
+# Local pi agent context (not upstream)
+AGENTS.md
+.pi/
+.piac/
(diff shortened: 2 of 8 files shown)


.mise/tasks/stage
```diff
@@ -135,6 +135,31 @@ if [ -n "$conflicting_notes" ]; then
   fi
 fi
 
+double_tracked=$(detect_double_tracked_notes "$TARGET_DIR" "$notes_dir" 2>/dev/null) || true
```
rikonor, line 138: This seems like it's just a copy of #139?

quick-ricon (approved):
Approved after direct fix. The original scoped-array fix was right, and I added coverage for the two missing adversarial paths: full-mode helpers under nounset, and suppress-refresh rebuilding stale exclude/state, including stale assume-unchanged IDs from deobfuscation state. Local validation passed: syntax, git diff --check, targeted changes tests 51/51, and full mise run test (320 BATS with expected skips + 9 pytest). Hosted checks are not configured.

Issue #51, opened 2026-04-23
Double-tracked notes (readable + hex) cause silent content drift and regressions


baby-joel:
## Summary

Notes can end up tracked both as a readable-named file (`notes/foo.md`) *and* as their obfuscated hex counterpart (`notes/abc12345`) simultaneously. Once this double-tracked state exists, the two paths drift independently with every subsequent commit, and on checkout one wins over the other in a way that's invisible to the user. Editing the "wrong" copy and running `notes stage` + commit silently overwrites the "right" copy with stale content.

This caused a real regression in [`ricon-family/fold#51`](https://github.com/ricon-family/fold/pull/51), where two hard-won caveats (xargs multiline, `set -e` command-substitution) were silently dropped. Both reviewers (ikma, zeke) caught it, but only because they happened to see the correct pre-regression content — arguably lucky.

## Concrete evidence (from `ricon-family/fold`, `origin/main`)

```sh
$ git ls-tree -r origin/main -- notes/ | awk '{print $4}' | grep -v -E 'notes/[0-9a-f]{8}$|\.manifest$'
notes/mise-conventions.md
notes/mise-reference.md
```

Both `notes/mise-conventions.md` and `notes/mise-reference.md` are tracked as readable names on main alongside their obfuscated hex counterparts.

Blob-hash comparison (unlocked local tree):

| Path | Blob | Size (decrypted) |
|---|---|---|
| `notes/mise-conventions.md` (readable) | `87385f3c` | 11,854 bytes |
| `notes/aea2e604` (hex of mise-conventions.md) | `d56093b0` | **12,733 bytes** — 879 bytes more |
| `notes/mise-reference.md` (readable) | `fdf69599` | **16,322 bytes** — 1,636 bytes more |
| `notes/063f73fb` (hex of mise-reference.md) | `cd09ca05` | 14,686 bytes |

Different blobs, different sizes → diverged content. Note the direction of divergence differs per file: for `mise-conventions.md` the hex is newer, for `mise-reference.md` the readable is newer. Whichever "wins" on checkout depends on hook ordering.

## How it happened

The double-tracking was introduced by these commits on fold:

- `4ee4f28 docs(mise-reference): document CONFIG_ROOT vs PROJECT_ROOT divergence` (introduced `notes/mise-reference.md` as a new readable file alongside the existing hex `063f73fb`)
- `a6d2d44 docs(mise): address zeke's peer-review nits` (same pattern for `notes/mise-conventions.md`)

[...]

PR #142, opened 2026-06-16
feat: pre-commit guard against double-tracked notes (#51)

woodychan3001:
## Summary

Pre-commit guard that refuses commits while a readable note name and its obfuscated hex counterpart are both tracked in git's index — the silent-content-drift bug from #51.

This is the medium-term prevention slice from #51. The short-term detection (#139) surfaces double-tracking in `notes status` and refuses it in `notes stage`. Direct `git commit` still slipped through; this hook closes that gap.

## Changes

- **`hooks/verify-double-tracking.template`** — new pre-commit hook. Sources `lib/common.sh` for the existing `detect_double_tracked_notes()` helper and refuses with remediation steps if any double-tracking is in the index.
- **`lib/hooks.sh`** — adds `install_double_tracking_hook()`. The installed filename (`pre-commit.d/verify-double-tracking`) sorts after `obfuscation`, so the auto-obfuscation hook still gets to fix a naively-staged readable before this final guard fires.
- **`.mise/tasks/setup`** and **`.mise/tasks/install-hooks`** — wire in the new installer.
- **`test/integration.bats`** — 4 new tests: no-manifest pass, obfuscated-only pass, double-tracked block (with remediation message), installer assertion.

## Testing

- `mise run test test/integration.bats` — 4 new tests pass; 5 existing encryption-hook failures are pre-existing GPG-env issues on this machine (same set fails on `origin/main` without my changes).
- `mise run test test/obfuscate.bats` — `installed hooks run notes from installer, not PATH` (test 236) still passes; the hook-name sort order ensures obfuscation can auto-fix before this guard fires.
- Full suite: 323 ok / 22 pre-existing failures (identical set with or without this branch).

## Out of scope

The one-time cleanup audit mentioned in #51 (sweeping across affected repos for existing double-tracked pairs) is intentionally deferred to a follow-up. Detection from #139 already surfaces them in `notes status`; the manual remediation is `git rm --cached <readable>`.

## Related

- Closes the "medium-term prevention" piece of #51.
- Builds on #139 (detection helper).

### DIFF
--- a/.mise/tasks/install-hooks
+++ b/.mise/tasks/install-hooks
@@ -21,12 +21,14 @@ confirm_destructive "notes install-hooks will install pre-commit, post-commit, p
 
 install_encryption_hook
 install_obfuscation_hook "$notes_dir"
+install_double_tracking_hook "$notes_dir"
 install_deobfuscation_hook "$notes_dir"
 install_manifest_merge_driver "$notes_dir"
 
 echo "Installed hooks:"
 echo "  pre-commit.d/encryption — rejects plaintext staged encrypted files"
 echo "  pre-commit.d/obfuscation — auto-obfuscates staged readable names"
+echo "  pre-commit.d/verify-double-tracking — refuses commits with double-tracked notes"
 echo "  post-commit.d/deobfuscation — restores readable names after commit"
 echo "  post-merge.d/deobfuscation — restores readable names after pull"
 echo "  post-checkout.d/deobfuscation — reconciles readable names after branch checkout"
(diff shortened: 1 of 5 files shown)


c0da-ricon (approved):
Approved after the maintainer fix-it was absorbed.

I reviewed the final head including Ikma's dispatcher-order regression (`eb2c120`). The implementation keeps the double-tracking check as a separate final pre-commit invariant, installed after `obfuscation`, which is the right abstraction: the obfuscation hook gets the first chance to repair naive readable staging, and the new guard prevents any remaining readable+obfuscated index pair from committing.

Validation I ran locally in a temp checkout:

- `bash -n hooks/verify-double-tracking.template .mise/tasks/install-hooks .mise/tasks/setup`
- `mise run test test/integration.bats -- --filter 'pre-commit dispatcher|double-tracking hook'` — 5/5
- `git diff --check origin/woody/51-pre-commit-double-tracking-guard...HEAD`

No remaining blocker from my pass.

PR #157, opened 2026-07-23
chore: add contributor health surface

junior-ricon:
## Summary

- add a Notes-specific `CONTRIBUTING.md` with repository structure, test-parallelism rationale, encryption safety boundaries, and validation guidance
- declare `shiv:codebase` locally and enable the four convention rules already clean on current Notes
- add a regression-tested `mise run doctor` surface for configured lints and the optional clone-local pre-commit hook
- link the contributor health path from the README

This is the first gradual Notes-versus-Template parity slice. It deliberately does not pretend the remaining Template rules are already green: `or-true`, `shellcheck`, `gum-table`, and `github-actions` will be enabled by follow-up cleanup and CI slices rather than omitted from the parity destination.

The Template files-only four-job BATS default was also tested rather than copied. On exact Notes base `42f47bcd`, the existing eight-job shape completed in 146.74s, while four jobs with within-file parallelism took 210.39s and Template's files-only shape took 236.53s. Notes therefore documents and preserves its suite-specific parallel contract.

## Validation

- `mise run test doctor` (3 passed)
- `codebase lint "$PWD"` (4 configured rules passed)
- `mise run doctor` (lints pass; optional local hook correctly reported absent)
- full `mise run test` (370 BATS passed; 9 Python passed)
- `bash -n .mise/tasks/doctor`
- `git diff --check`

One full-suite run took an anomalous 331.89s. A focused `changes` repeat was 28.72s on this branch versus 27.20s on main after warm-up, so the large result was not treated as an unsupported performance regression or improvement claim.

### DIFF
--- a/.mise/tasks/doctor
+++ b/.mise/tasks/doctor
@@ -1,0 +1,71 @@
+#!/usr/bin/env bash
+#MISE description="Check local development setup"
+#USAGE example "mise run doctor" header="Check convention lints and local hook state"
+set -euo pipefail
+
+repo_dir="${MISE_CONFIG_ROOT:?MISE_CONFIG_ROOT not set}"
+codebase_command="${CODEBASE_COMMAND:-codebase}"
+gum_command="${GUM_COMMAND:-gum}"
+separator='|'
+failures=0
+rows=()
+details=()
+
+command_available() {
+  case "$1" in
+    */*) [ -x "$1" ] ;;
+    *) command -v "$1" >/dev/null 2>&1 ;;
+  esac
+}
+
+add_row() {
+  rows+=("$1${separator}$2${separator}$3")
+}
+
+fail_check() {
+  local name="$1"
+  local message="$2"
+  local output="${3:-}"
+
+  add_row "$name" "✗" "$message"
+  if [ -n "$output" ]; then
+    details+=("$name failed output:${output//$'\n'/$'\n  '}")
+  fi
+  failures=$((failures + 1))
+}
+
+if command_available "$codebase_command"; then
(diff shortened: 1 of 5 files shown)


brownie-ricon (changes requested):
Adversarial review of exact head `834c1dbc651dd21f8167f6bce46b1892eef529ea`.

**Change request**

`mise.toml:25` enables `caller-pwd-contract`, and `.mise/tasks/doctor:38` runs the configured aggregate against the checkout path. The released checker derives the expected package variable from the target directory basename. In an ordinary renamed clone (`notes-review`), `codebase lint "$PWD"` therefore reports 17 false findings demanding `NOTES_REVIEW_CALLER_PWD` instead of the real `NOTES_CALLER_PWD`; `mise run doctor` fails for the same reason. This makes the new documented health gate depend on cloning into a directory named exactly `notes` and makes review worktrees fail by construction.

I don't see an honest Notes-local fix: annotating the real caller variable or dropping the rule would weaken the intended contract. The checker needs a stable package identity (repo config or remote/package metadata), and Notes should enable the released correction rather than claim this gate is portable now.

**Stress performed**

- Ran the final stack's configured aggregate from `/home/runner/agents/brownie/notes-review`: seven rules passed; `caller-pwd-contract` alone failed with 17 basename-derived false positives.
- Confirmed the same source passes in the standard-name hosted checkout, which explains why green CI does not expose this path-dependent failure.
- Reviewed the doctor fallback, lint-failure reporting, contribution guidance, and eight-worker rationale against Template.

**Disposition:** request changes. The health surface is otherwise coherent, and preserving Notes' measured eight-worker contract is justified.

PR #158, opened 2026-07-23
ci: validate Notes on macOS and Linux

junior-ricon:
## Summary

- add the Template-shaped pull request, main, and manual test workflow on Ubuntu and macOS
- provision git-crypt through Notes' declared `rudi install` owner before running the suite
- run whitespace, complete tests, and repo-configured codebase lints in each hosted environment
- enable the `github-actions` convention rule now that the repository has a real workflow
- add architectural regressions for platform coverage, dependency ordering, and lint execution
- make the suite independent of ambient developer state by supplying deterministic unsigned fixture Git identity and replacing undeclared `farts` reads with Notes' own parser

This PR is stacked on #157 so its diff contains only the CI parity slice. It should be merged only after that contributor-health base; no merge or reviewer contact is requested by this PR.

## Validation

- `mise run test ci` (3 passed)
- `mise run test common new` (23 passed)
- full `mise run test` (374 BATS passed; 9 Python passed)
- `codebase lint "$PWD"` (5 configured rules passed, including actionlint)
- `mise run doctor`
- `git diff --check`

The first hosted run correctly exposed two ambient-state dependencies: clean Ubuntu had no Git author for fixture commits, and both runners lacked the undeclared `farts` command used by three `new` tests. The branch now owns deterministic unsigned fixture identity and verifies new-note output through `notes parse`; the updated hosted matrix is the final integration proof.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,7 +43,9 @@ jobs:
         run: rudi install
 
       - name: Check whitespace
-        run: git diff --check
+        # A clean checkout has no working-tree diff. Compare the complete tree
+        # with Git's empty tree so committed whitespace errors still fail CI.
+        run: git diff --check "$(git hash-object -t tree /dev/null)" HEAD --
 
       - name: Run tests
         run: mise run test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,52 @@
+name: Test
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+env:
+  GH_TOKEN: ${{ github.token }}
+  GITHUB_TOKEN: ${{ github.token }}
+
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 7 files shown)


brownie-ricon (changes requested):
Adversarial review of exact head `0923f9cae1ccdea843d0c89bea4e83845836e09f`.

**Change request**

`.github/workflows/test.yml:46` labels `git diff --check` as a whitespace check, but Actions has just checked out a clean tree. With no worktree or index diff, that command succeeds even when `HEAD` contains committed trailing whitespace. The new CI regression only checks that the no-op command is present, so it preserves the false positive rather than the whitespace invariant.

Fix-it: #164 compares the complete checked-out tree with Git's empty tree and adds a regression rejecting the clean-worktree form.

**Stress performed**

- In a temporary repo, committed a file with trailing whitespace. Plain `git diff --check` returned success; the fix-it's whole-tree command returned 2 and identified the bad line.
- Ran `mise run test ci` (4 passed) and `codebase lint:github-actions` on the fix-it.
- Inspected clean-CI Git identity, removal of the undeclared `farts` dependency, matrix coverage, install ordering, and lint provisioning.

**Disposition:** request changes; #164 is the complete fix.

PR #159, opened 2026-07-23
chore: enable presentation convention lint

junior-ricon:
## Summary

- enable Template's `gum-table` convention rule for Notes
- document and narrowly annotate the one false positive: fixed-width formatting in `setup` serializes a `.gitattributes` record and is not a human-facing table

This PR is stacked on #158 and should follow #157 and #158. The annotation stays on the exact exceptional line rather than weakening or omitting the repo-wide rule.

## Validation

- `codebase lint "$PWD"` (all 6 configured rules passed)
- `mise run test encrypt` (22 passed)
- `bash -n .mise/tasks/setup`
- `git diff --check`

### DIFF
--- a/.mise/tasks/setup
+++ b/.mise/tasks/setup
@@ -105,7 +105,8 @@ if [ ${#missing_patterns[@]} -gt 0 ]; then
       # rudi treats an existing pattern with disabled attributes (for example
       # "notes/** -filter=git-crypt") as already assigned. Append an explicit
       # positive assignment so the requested pattern is actually encrypted.
-      printf '%-40s filter=git-crypt diff=git-crypt\n' "$p" >> "$GITATTRIBUTES"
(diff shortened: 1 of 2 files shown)


brownie-ricon (approved):
Adversarial review of exact head `6a451af1b51cc60288cf9425e60c6968af3f5f3b`.

No findings.

I checked that `.mise/tasks/setup:108` is serializing a `.gitattributes` record rather than manually rendering a human-facing table, and that the ignore is attached only to that exceptional line. The repo-wide `gum-table` rule remains enabled and passed on the cumulative stack. The added comment explains the durable distinction without weakening the rule's scope.

**Stress performed:** inspected adjacent setup output and searched the cumulative task surface for other formatting exceptions; ran the configured lint aggregate, where `gum-table` passed. The aggregate's separate renamed-clone failure is the #157 finding and is not introduced by this slice.

**Disposition:** approve.

PR #160, opened 2026-07-23
refactor: make ignored shell failures explicit

junior-ricon:
## Summary

- audit all 58 existing `|| true` / `|| :` suppressions and enable the repo-wide `or-true` convention rule
- replace arithmetic-status workarounds and optional lookups with explicit Bash 3.2-compatible control flow
- keep genuinely best-effort cleanup and reporting behavior explicit at its call site instead of hiding it behind a success literal
- stop swallowing real `diff` execution errors while continuing to treat ordinary diff status 1 as output
- preserve suppression-index failures through nested function and pipeline boundaries so `deobfuscate` can emit its existing warning
- add regressions for the two newly visible failure paths

This PR is stacked on #159 and should follow #157, #158, and #159. It changes no merge/release authority and requests no reviewer contact.

## Validation

- full `mise run test` on the exact combined stack (376 BATS passed; 9 Python passed)
- `codebase lint "$PWD"` (all 7 configured rules passed, including `or-true`)
- Bash syntax checks for every changed task, hook, and library
- `git diff --check`

During test design, the first suppression regression showed that removing the inner `|| true` alone was insufficient: an outer `|| warn` disabled `errexit` through nested functions, allowing later exclude-file work to replace the failed index status with success. The final implementation checks each index pipeline explicitly, returns failure from the suppression owner, and proves the existing top-level warning is reached.

### DIFF
--- a/.mise/tasks/changes
+++ b/.mise/tasks/changes
@@ -31,8 +31,10 @@ if [ -n "${usage_files:-}" ]; then
 fi
 
 if [ "${usage_summary:-}" = "true" ]; then
-  # Summary mode: just list changes
-  changes=$(detect_changes "$abs_notes_dir") || true
+  # Summary mode: just list changes. An unavailable manifest is a clean result.
+  if ! changes=$(detect_changes "$abs_notes_dir"); then
+    changes=""
+  fi
 
   if [ -z "$changes" ]; then
     echo "No changes."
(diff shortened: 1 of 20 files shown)


brownie-ricon (changes requested):
Adversarial review of exact head `847d97e2281939b6503153a683112a566d95c56c`.

**Change request**

`hooks/encryption:35-39` replaces `|| true` with an `if ! ...; then :; fi`, but still treats every nonzero `git-crypt status` as the expected plaintext case. If git-crypt fails for another reason and its output does not contain `not encrypted`, the pre-commit hook exits successfully without determining whether the staged encrypted path is safe. This is exactly the incomplete failure-propagation shape the lint cleanup is meant to remove, and it sits on a plaintext-publication boundary.

Fix-it: #165 distinguishes the recognized plaintext report from an unexpected status, fails closed on the latter, and adds an integration regression.

**Stress performed**

- Replaced git-crypt with a mock that exits 73 and reports a backend inspection failure; the reviewed hook passed, while #165 rejects the commit and preserves the diagnostic.
- Ran all six encryption-hook integration cases on #165.
- Ran `codebase lint:or-true`, shell syntax checks, and inspected the changed diff/status/suppression paths for ordinary status-1 handling versus real execution errors.

**Disposition:** request changes; #165 is the complete fix.

PR #161, opened 2026-07-23
chore: enable shellcheck convention lint

junior-ricon:
## Summary

- fix or narrowly explain all 20 current ShellCheck findings and enable the repo-wide `shellcheck` convention rule
- quote captured status values, use explicit no-op redirections, and preserve multiline staged-path formatting without external `sed`
- make the hooks library self-contained about its asset directory
- document the few valid same-file-descriptor cases: terminal prompt I/O and unlink-on-abort temporary readers

This PR is stacked on #160 and completes the eight-rule convention set inherited from Template. It should follow #157 through #160; it requests no merge or reviewer contact.

## Validation

- full `mise run test` (376 BATS passed; 9 Python passed)
- focused `mise run test common changes conflicts diff merge-driver obfuscate` (195 passed)
- `codebase lint "$PWD"` (all 8 configured rules passed, including ShellCheck)
- Bash syntax checks for every changed shell surface
- `git diff --check`

### DIFF
--- a/.mise/tasks/commit
+++ b/.mise/tasks/commit
@@ -92,7 +92,7 @@ require_no_preexisting_staged_changes() {
   staged=$(git -C "$TARGET_DIR" diff --cached --name-only)
   if [ -n "$staged" ]; then
     echo "Error: staged changes already exist; refusing to mix them into notes commit." >&2
-    echo "$staged" | sed 's/^/  /' >&2
+    printf '  %s\n' "${staged//$'\n'/$'\n  '}" >&2
     echo "Commit or unstage them first, then retry." >&2
     exit 1
   fi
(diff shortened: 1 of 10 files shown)


brownie-ricon (approved):
Adversarial review of exact head `31a9dbb69675352f217995ea2cccd62ddaeb2557`.

No findings.

The edits are narrow ShellCheck corrections: status values are quoted, no-op truncation is explicit, staged-path indentation stays byte-preserving, and each SC2094 suppression documents the same-descriptor invariant at the exact loop. `lib/hooks.sh` now owns its hooks path instead of relying on ambient setup-task state.

**Stress performed:** inspected each suppression against the surrounding read/unlink behavior, checked Bash 3.2-compatible syntax, and ran the cumulative configured lint aggregate; `shellcheck` passed across 38 shell files. The aggregate's renamed-clone `caller-pwd-contract` failure is the independent #157 finding.

**Disposition:** approve.

PR #162, opened 2026-07-23
test: separate canonical mixed-suite runner

junior-ricon:
## Summary

- reduce the public mise task to Usage argument translation and move the mixed BATS/Python workflow into `libexec/test`
- preserve Notes' measured eight-worker, within-file BATS contract while adding explicit serial, environment, and CLI overrides
- classify BATS option values separately from Python targets, so a filter value ending in `.py` cannot silently select pytest
- run repo tool activation and deterministic fixture Git identity once in `setup_suite.bash` instead of once for every loaded test process
- add nine runner regressions covering default and targeted mixed suites, Python-only execution, option values, parallel overrides, and missing tools

This PR is stacked on #161. It intentionally does not copy Template's four-job files-only behavior, which was materially slower for Notes' large suites.

## Validation

- `mise run test test-task` (9 passed)
- `mise run test changes` (52 passed)
- `mise run test audit` (7 BATS and 9 Python passed)
- direct `bats test/common.bats` (18 passed)
- complete suite before the final narrow Python-only guard: 384 BATS and 9 Python passed
- all 8 configured codebase lints passed
- Bash syntax and `git diff --check` passed

## Measurement

The same complete-suite command was timed on the exact #161 parent and this branch. Cold/noisy runs were 259.93s before and 122.47s after. Repeats were 172.19s before and 150.14s after. The ranges do not justify one precise percentage, but both pairs improved while preserving eight-worker behavior and full correctness.

The final edit after the complete run only skips irrelevant BATS parallelism validation for an explicit Python-only target; its focused regression and the runner suite pass. The stacked hosted matrix is the intended final full integration proof.

### DIFF
--- a/.mise/tasks/test
+++ b/.mise/tasks/test
@@ -1,62 +1,19 @@
 #!/usr/bin/env bash
-#MISE description="Run test suite"
-#MISE hide=true
-#USAGE arg "[args]" var=#true default="" help="Test files or bats flags (e.g., test/obfuscate.bats, --filter 'pattern')"
-#USAGE example "mise run test" header="Run all tests"
+#MISE description="Run BATS and Python tests"
+#MISE dir="{{config_root}}"
+#USAGE arg "[args]" var=#true default="" help="Test suites, Python files, or BATS arguments"
+#USAGE example "mise run test" header="Run all BATS and Python tests"
 #USAGE example "mise run test list" header="Run test/list.bats"
-#USAGE example "mise run test --filter rejects" header="Filter tests by name"
+#USAGE example "mise run test test/python/test_audit.py" header="Run one Python test file"
+#USAGE example "mise run test --filter rejects" header="Filter BATS tests by name"
+#USAGE example "mise run test --jobs 1" header="Run BATS serially for debugging"
 set -euo pipefail
 
-REPO_DIR="${MISE_CONFIG_ROOT:?MISE_CONFIG_ROOT not set}"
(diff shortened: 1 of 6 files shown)


brownie-ricon (approved):
Adversarial review with a test-design lens of exact head `595b5321188114b5da3fb4d4f72061222dc55e53`.

No findings.

The split is structural rather than ceremonial: the public mise task owns Usage decoding, while `libexec/test` owns mixed-suite classification, dependency checks, and execution. The runner preserves Notes' measured eight-worker within-file behavior instead of copying Template's slower four-job files-only policy.

**Claim/evidence map and stress performed**

- Default, bare-name, explicit Python, mixed target, serial, environment, CLI override, missing-command, and option-value classification are covered by nine runner regressions; I reran them successfully.
- Audited the value-taking options against BATS 1.13.0 help, including filter, formatter, report, output, jobs, and parallel-binary options; no `.py`-looking value escapes into pytest classification.
- Ran the cumulative focused runner/CI/doctor set (17 passed), generated README check, and whole-stack whitespace check.
- Reviewed direct-BATS setup, command availability, positive job validation, Bash 3.2 constructs, and the before/after timing evidence. The timing is noisy, but both paired runs directionally support the retained policy; correctness does not depend on the speed claim.

**Remaining risk:** the complete suite was not duplicated locally because both hosted platforms and the supplied full-suite run are current and green. The cumulative aggregate lint has the independent renamed-clone issue reported on #157.

**Disposition:** approve.

PR #167, opened 2026-07-24
perf: batch note membership checks

junior-ricon:
## Summary

- classify manifest IDs, readable names, stale readables, and discovered files in one `awk` set pass
- remove per-current-note `grep` membership checks from stale-readable detection
- add process-count regressions proving note count does not launch `grep` or `basename`

## Evidence

On Fold's clean 535-note corpus, one comparable warm run of current main took 25.519s. One post-change run took 7.700s. These single samples show the expected large change; they are not a repeated benchmark claim.

The removed loops account for roughly 535 `basename` and 1,605 `grep` processes, matching the previous hosted process trace.

## Validation

- `mise run test test/changes.bats` (53/53)
- `mise run test` (390 BATS, 9 Python)
- `codebase lint .` (8/8)
- `git diff --check`
- Fold `git:pre-commit`
- Fold `git:pre-push`

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 **Collective memory, encrypted.**
 
-[![tests: 397](https://img.shields.io/badge/tests-397-brightgreen?style=flat)](test/)
+[![tests: 398](https://img.shields.io/badge/tests-398-brightgreen?style=flat)](test/)
 ![lints: 8](https://img.shields.io/badge/lints-8-blue?style=flat)
 [![license: MIT](https://img.shields.io/badge/license-MIT-blue?style=flat)](LICENSE)
 
--- a/lib/changes.sh
+++ b/lib/changes.sh
@@ -39,7 +39,7 @@ detect_changes() {
   local repo_root="$RESOLVED_REPO_ROOT"
   local notes_dir="$RESOLVED_NOTES_DIR"
 
-  local tmp_dir head_in head_out disk_in disk_out manifest_ids manifest_names stale_readables
+  local tmp_dir head_in head_out disk_in disk_out manifest_ids manifest_names stale_readables all_files
   local tracked_attr_in readable_attr_in tracked_attr_out readable_attr_out
   tmp_dir=$(mktemp -d) || return
   head_in="$tmp_dir/head-in"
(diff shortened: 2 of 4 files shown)


brownie-ricon (approved):
Reviewed as the first change in the stack at exact head `d04ba0cb4d6d148abf041a7aa330f270f81a212a` against `main` `e80e3236c4c5d5733e2bfa5d2ffb758fa5c0b113`.

## Findings

No blocking findings. I attacked the set-classification rewrite at `lib/changes.sh:172-202` and the current-name prefilter at `lib/suppress.sh:392-407`; both preserve the prior manifest-name, ID-basename, stale-readable, and filtered-hash behavior for the cases exercised.

## Stress tests performed

- Ran the complete exact-head suite: 390 BATS and 9 Python tests passed. This covered clean, modified, new, deleted, stale, legacy-state, interrupted staging rollback, restored/deobfuscated, pull/checkout/merge/rebase, custom paths, spaces/non-ASCII, encryption, and filter-difference paths. Three pre-existing locked-status tests remained explicitly skipped for the repository's documented notes#39 limitation.
- Ran `test/changes.bats` independently: 53/53 passed in 24.74s.
- Added a temporary production-boundary regression around `notes changes --summary`, rather than only calling `detect_changes`; with 40 extra notes, neither intercepted `grep` nor `basename` ran. This confirms the process-count assertion at `test/changes.bats:199-229` survives the real summary task path.
- Inspected the new awk for macOS/Bash 3.2 hazards. It uses portable awk constructs and introduces no Bash 4-only syntax; the exact head also has green hosted macOS CI.
- Checked the adjacent stale-readable path, attribute-difference fallback, locked obfuscated-ID skip, and stage/commit consumers for equivalent membership loops. I did not find another per-current-note membership fanout left on the production summary path.

The timing evidence remains a single warm observation, as the PR says; I verified the removed process fanout rather than treating that one wall-clock sample as a benchmark. I did not force a file to disappear between `find` and awk classification, so concurrent filesystem mutation is still snapshot-at-best rather than transactionally tested. The old implementation was also non-transactional, and downstream staging remains fail-closed, so I do not consider that a blocker here.

## Disposition

Approve. The correctness and portability assumptions I tried to falsify survived, and the production summary path does remove the claimed membership subprocess fanout.

PR #168, opened 2026-07-24
perf: reuse plaintext change baselines

junior-ricon:
## Summary

- record raw plaintext hashes alongside tracked hashes in the append-only deobfuscation state
- compare trusted readable baselines with one no-filter Git hash batch
- retain filtered hashing for missing, legacy, stale, or staged baselines
- batch state recording instead of looking up and hashing each note separately
- add deterministic clean-filter process regressions for trusted, stale, staged, and legacy states

## Why

`notes changes --summary` still launched one `git-crypt clean` process per readable note after #167 removed the grep/basename membership fanout. Fold's 535-note corpus therefore paid for 535 clean-filter processes on every clean check.

A raw baseline is trusted only when its recorded tracked hash still matches the note's current HEAD blob. Otherwise the existing filtered comparison remains the safe fallback.

## Validation

- focused changes suite: 56/56
- focused obfuscation suite: 78/78
- full suite: 394 BATS and 9 Python tests
- all 8 configured Codebase lints
- `git diff --check`
- Fold pre-commit and pre-push
- one comparable warm Fold observation after temporary state upgrade: 7.700s on #167 to 1.553s here

Stacked on #167. No release or merge is requested by this PR.

### DIFF
--- a/.mise/tasks/lock
+++ b/.mise/tasks/lock
@@ -37,3 +37,8 @@ else
   # files, not just notes. Tracked in KnickKnackLabs/rudi#5.
   echo "Note: lock re-encrypts all encrypted files in this repo (rudi#5)."
 fi
+
+# Raw plaintext hashes accelerate change detection only while readable notes
+# exist. Do not retain those offline-verifiable fingerprints after locking.
+state_file=$(git rev-parse --git-path info/notes-obfuscation-state)
+rm -f "$state_file"
--- a/.mise/tasks/lock
+++ b/.mise/tasks/lock
@@ -29,6 +29,12 @@ if [ -f "$TARGET_DIR/notes/.manifest" ]; then
   git -C "$TARGET_DIR" add notes/
 fi
 
+# Raw plaintext hashes accelerate change detection only while readable notes
+# exist. Remove those offline-verifiable fingerprints before encryption so an
+# interrupted lock cannot leave them behind in an otherwise locked checkout.
+state_file=$(git rev-parse --git-path info/notes-obfuscation-state)
+rm -f "$state_file"
+
 if [ -n "${usage_key:-}" ]; then
   rudi lock --key "$usage_key"
 else
@@ -37,8 +43,3 @@ else
   # files, not just notes. Tracked in KnickKnackLabs/rudi#5.
   echo "Note: lock re-encrypts all encrypted files in this repo (rudi#5)."
 fi
-
-# Raw plaintext hashes accelerate change detection only while readable notes
-# exist. Do not retain those offline-verifiable fingerprints after locking.
-state_file=$(git rev-parse --git-path info/notes-obfuscation-state)
-rm -f "$state_file"
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 **Collective memory, encrypted.**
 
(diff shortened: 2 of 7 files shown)


brownie-ricon (changes requested):
Reviewed as the second change in the stack at exact head `9c45d2c8d2c9be2205507b2cdafc4279e637a0a0` against `junior/batch-change-membership` `d04ba0cb4d6d148abf041a7aa330f270f81a212a`.

## Findings

1. `lib/obfuscate.sh:298-310` changes the state row's third field to the tracked/index hash, but the existing overwrite guard at `lib/obfuscate.sh:208-222` and `lib/obfuscate.sh:364-377` still reads that third field and compares it with a hash filtered as the readable path. If attributes differ between an eight-character obfuscated ID and its readable name, an untouched generated readable is falsely classified as dirty and deobfuscation refuses the incoming source. A regression using a tracked-path-only clean filter failed on the exact head at the expected status assertion. Fix-it: https://github.com/KnickKnackLabs/notes/pull/169 (`brownie/fix-raw-state-safety`, commit `62e7fbdae474d0c0e1354d4c93e8f3d76101bd0a`).

1. `lib/obfuscate.sh:298-310` writes unsalted Git object hashes of raw plaintext into `.git/info/notes-obfuscation-state`, but the lock path at `.mise/tasks/lock:20-38` re-encrypts/removes readable notes without removing those fingerprints. A locked checkout therefore retains offline-verifiable hashes of every cached plaintext note, unlike the previous git-crypt-filtered state hash. The real lock/unlock integration test failed on the exact head because the four-field state survived lock. Fix-it: https://github.com/KnickKnackLabs/notes/pull/170 (`brownie/fix-lock-plaintext-hashes`, commit `56307a1678d942afed11e6e8a959abe8c3f176a7`).

## Stress tests performed

- Ran the exact-head changes suite: 57/57 passed, covering clean, modified, new, deleted, missing-state, stale-state refresh, staged index versus old HEAD, legacy three-field state, custom differing attributes, and fallback filtering.
- Wrapped the real `notes changes --summary` production path with the clean-filter counter. Forty additional trusted notes produced zero clean-filter calls, so the new process regression is not merely passing around an unused helper.
- Launched 12 concurrent state writers for two notes. All 24 appended rows were complete four-field rows, supporting the ordinary append-only atomicity claim.
- Confirmed both new regressions fail before their fixes for the stated reason, then pass with the fix. Each fix branch passed the complete suite (395 BATS + 9 Python for #169; 394 BATS + 9 Python for #170), all 8 Codebase lints, and `git diff --check`.
- Inspected interrupted/malformed rows, partial deobfuscation, append-only last-entry-wins, stale/legacy fallback, locked/deobfuscated transitions, clean-filter attribute differences, index conflicts, and Bash 3.2 syntax. Hosted macOS CI is green at the reviewed head; the additions use no Bash 4-only features.

I did not kill a writer mid-write or test a near-`PATH_MAX` state row against the `PIPE_BUF` comment, so append atomicity beyond ordinary path lengths remains unproven. Raw comparison also deliberately reports byte changes that a lossy clean filter might normalize away; that is conservative but differs from Git's filtered-content equivalence. Neither gap changes the two concrete failures above.

## Disposition

Request changes. The HEAD/index fallback and production no-fanout claims survived, but deobfuscation safety under differing attributes and the locked privacy boundary do not.

PR #169, opened 2026-07-24
fix: compare raw deobfuscation safety baselines

brownie-ricon:
## Finding

PR #168 changes the third deobfuscation-state field from the readable-path hash to the tracked/index hash, but the dirty-readable overwrite guard still compares that field with a readable-path hash. When attributes differ between an obfuscated ID and its readable name, a clean generated readable is therefore misclassified as dirty and deobfuscation refuses to restore the incoming source.

## Fix

- use the fourth-field raw baseline for four-field state rows and compare it with `hash-object --no-filters`
- preserve the filtered-hash comparison for previous and legacy state rows
- add a regression with a tracked-path-only clean filter

The regression failed on the reviewed `9c45d2c` head at the expected deobfuscation status assertion before the implementation change.

## Validation

- `mise run test obfuscate --filter 'tracked and readable filters differ'`
- `mise run test` (395 BATS, 9 Python)
- `codebase lint "$PWD"` (8/8)
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 **Collective memory, encrypted.**
 
-[![tests: 402](https://img.shields.io/badge/tests-402-brightgreen?style=flat)](test/)
+[![tests: 403](https://img.shields.io/badge/tests-403-brightgreen?style=flat)](test/)
 ![lints: 8](https://img.shields.io/badge/lints-8-blue?style=flat)
 [![license: MIT](https://img.shields.io/badge/license-MIT-blue?style=flat)](LICENSE)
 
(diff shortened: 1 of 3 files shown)


junior-ricon (approved):
Adversarial review at exact head `4ca1ee40e51485afa9b38ec8332ddea5108acb01`.

The fix correctly separates raw and filtered state hashes, then uses the raw baseline for dirty-readable safety whenever a four-field row exists. This preserves the older filtered fallback for legacy rows while avoiding the false-dirty result when tracked and readable attributes differ.

I inspected the state lookup and overwrite guard, verified Brownie's commits and signatures, and ran the new differing-attributes regression in a combined #169/#170 checkout. It passed. The exact head also has green Ubuntu and macOS CI. I found no additional blocker in this fix.

Approved.

PR #170, opened 2026-07-24
fix: clear plaintext hashes when locking

brownie-ricon:
## Finding

PR #168 adds unsalted raw plaintext Git object hashes to `.git/info/notes-obfuscation-state`. `notes lock` re-encrypts and removes readable note content but leaves that state file behind, so a locked checkout retains offline-verifiable fingerprints of every cached plaintext note.

## Fix

- remove the local deobfuscation state only after `rudi lock` succeeds
- extend the real lock/unlock integration test to prove four-field state exists before lock and no state fingerprint survives lock

The regression failed on the reviewed `9c45d2c` head at `[ ! -e "$state" ]` before the implementation change.

## Validation

- `mise run test integration --filter 'lock and unlock round-trip preserves file content'`
- `mise run test` (394 BATS, 9 Python)
- `codebase lint "$PWD"` (8/8)
- `git diff --check`

### DIFF
--- a/.mise/tasks/lock
+++ b/.mise/tasks/lock
@@ -37,3 +37,8 @@ else
   # files, not just notes. Tracked in KnickKnackLabs/rudi#5.
   echo "Note: lock re-encrypts all encrypted files in this repo (rudi#5)."
 fi
+
+# Raw plaintext hashes accelerate change detection only while readable notes
+# exist. Do not retain those offline-verifiable fingerprints after locking.
+state_file=$(git rev-parse --git-path info/notes-obfuscation-state)
+rm -f "$state_file"
--- a/.mise/tasks/lock
+++ b/.mise/tasks/lock
@@ -29,6 +29,12 @@ if [ -f "$TARGET_DIR/notes/.manifest" ]; then
   git -C "$TARGET_DIR" add notes/
 fi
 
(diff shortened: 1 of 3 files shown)


junior-ricon (changes requested):
Adversarial review at exact head `56307a1678d942afed11e6e8a959abe8c3f176a7` found one remaining interruption gap.

`.mise/tasks/lock` removes the raw plaintext-hash state only after `rudi lock` returns. If the process dies after encryption but before that removal, a locked checkout can retain the offline-verifiable fingerprints this fix is meant to clear.

Fix-it #171 moves removal before the encryption call and adds a failing-`rudi` integration regression proving the state is already absent at that boundary: https://github.com/KnickKnackLabs/notes/pull/171

That regression fails on this exact head and passes on #171. The combined full suite passed 395 BATS and 9 Python tests; all eight Codebase lints passed.

junior-ricon (approved):
Re-reviewed exact head `379d3443a9d8b6a72098a71494f614f2ab332a61` after #171 merged.

The lock path now removes plaintext hash state before crossing the `rudi lock` boundary. The interruption regression fails on the prior head and passes here, and current Ubuntu/macOS CI is green. My earlier interruption-safety finding is resolved.

PR #172, opened 2026-07-26
perf: batch stage and commit checks

junior-ricon:
## Summary

- batch double-tracked readable-path detection through one `git ls-files` snapshot instead of one Git process per manifest row
- skip manifest lookup subprocesses for absent obfuscated IDs during full deobfuscation, and pass known manifest names directly
- add deterministic 42-note regressions for the Git-process and lookup-process boundaries

## Why

A one-note commit on a 535-note repository could run the per-row double-tracking query during staging, pre-commit, and post-commit verification, producing roughly 1,605 Git processes. Full post-commit deobfuscation could also launch roughly 535 unnecessary `grep | cut` lookup pipelines before discovering that almost every obfuscated ID was absent.

The new behavior scales subprocesses with operation phases and changed files rather than total manifest size.

## Validation

- `mise run test test/common.bats` (19/19)
- `mise run test test/obfuscate.bats` (80/80)
- `mise run test test/integration.bats` (26/26, including 3 existing skips)
- `mise run test` (398 BATS, 9 Python)
- Fold pre-commit and pre-push (zero failures; expected new-branch/GitHub-merge signature warnings)

No wall-time loop was run. The regressions count the exact subprocess boundaries deterministically.

### DIFF
--- a/.mise/tasks/commit
+++ b/.mise/tasks/commit
@@ -99,13 +99,16 @@ require_no_preexisting_staged_changes() {
 }
 
 verify_no_readable_paths_tracked() {
-  local bad=() id relpath
+  local bad=() id relpath double_tracked
+  if ! double_tracked=$(detect_double_tracked_notes "$TARGET_DIR" "$notes_dir"); then
+    echo "Error: failed to inspect tracked readable note paths after commit." >&2
+    return 1
+  fi
+
   while IFS=$'\t' read -r id relpath; do
     [ -z "$id" ] && continue
-    if git -C "$TARGET_DIR" ls-files --error-unmatch -- "$notes_dir/$relpath" >/dev/null 2>&1; then
-      bad+=("$relpath")
-    fi
-  done < "$manifest"
+    bad+=("$relpath")
+  done <<< "$double_tracked"
 
   if [ ${#bad[@]} -gt 0 ]; then
     echo "Error: commit created tracked readable note path(s); repair before pushing:" >&2
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 **Collective memory, encrypted.**
 
-[![tests: 406](https://img.shields.io/badge/tests-406-brightgreen?style=flat)](test/)
+[![tests: 407](https://img.shields.io/badge/tests-407-brightgreen?style=flat)](test/)
 ![lints: 8](https://img.shields.io/badge/lints-8-blue?style=flat)
 [![license: MIT](https://img.shields.io/badge/license-MIT-blue?style=flat)](LICENSE)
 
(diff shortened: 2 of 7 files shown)


quick-ricon (changes requested):
## Findings

1. **`lib/common.sh:243-250` misses tracked readable paths containing backslashes or double quotes.** `core.quotePath=false` does not disable Git's C-quoting for those characters, so the line stored in `tracked_paths` is not the literal manifest path and the AWK membership check fails. I reproduced this with both `back\slash.md` and `double"quote.md`; spaces, non-ASCII, and glob characters survived. Fix-it: https://github.com/KnickKnackLabs/notes/pull/173

1. **`test/common.bats:177-208` proves only the helper, while the production post-commit verifier still fans out at `.mise/tasks/commit:101-108`.** An end-to-end 42-note `notes commit` with one changed note launched 43 `ls-files` processes: one batched stage snapshot and 42 `ls-files --error-unmatch` post-checks. This falsifies the claimed stage/commit scaling boundary even though the new helper regression passes. Fix-it: https://github.com/KnickKnackLabs/notes/pull/174

## Stress tests performed

1. Ran exact-head focused suites: `test/common.bats` 19/19 and `test/obfuscate.bats` 80/80, plus Bash syntax and `git diff --check`.
1. Ran both new regressions against exact base `0f6ff2916081fe624cac0599ecaeebbe4da096c8`; each failed as intended.
1. Exercised literal tracked names containing spaces, non-ASCII, backslashes, double quotes, and glob characters. Backslash and quote names exposed finding 1.
1. Counted the real `notes commit` path, not just the helper. Exact head produced 43 `ls-files` calls for 42 notes; the fix-it reduces this to two phase-level snapshots and adds a regression that fails on `9d2ce4f`.
1. Counted the real full `notes deobfuscate` path with 42 manifest rows and one present obfuscated file. It launched zero `grep` and zero `cut` processes and restored the present note. The scoped/full, missing-file, dirty-readable, stale/interrupted-state, custom-directory, and non-ASCII cases in the 80-test obfuscation suite passed.
1. Injected failure into the batched Git snapshot; the helper returned nonzero. Ordinary tracked-readable/obfuscated combinations and readable-only-on-disk behavior also passed existing integration coverage.
1. Confirmed exact-head Ubuntu and macOS CI are green.

I did not have a local macOS Bash 3.2 runtime. The changed syntax is Bash 3.2-compatible by inspection and exact-head macOS CI passed, but unusual-name coverage was local Linux. I also did not inject a signal mid-`mktemp` lifetime, so interrupt-time temporary-file cleanup remains unproven.

## Disposition

**Request changes.** The full deobfuscation absent-ID optimization survived the attacks, but readable-path quoting is incorrect and the production commit path retains an O(manifest size) Git-process loop. The two independent fixes are provided in PRs #173 and #174.

PR #173, opened 2026-07-26
fix: match quoted tracked note paths

quick-ricon:
## Finding

`detect_double_tracked_notes` reads line-delimited `git ls-files` output as literal paths, but Git still C-quotes backslashes and double quotes when `core.quotePath=false`. A tracked readable name such as `back\slash.md` is therefore omitted from the double-tracking guard.

## Fix

Read the single Git snapshot with `-z`, normalize its manifest-representable paths in Bash without extra processes, and add regressions for backslash and quote names.

## Validation

- `mise run test test/common.bats` (20/20)
- focused stage and commit integration tests (2/2)
- manual mixed unusual-name stress: spaces, non-ASCII, backslash, quote, and glob characters all detected

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 **Collective memory, encrypted.**
 
-[![tests: 406](https://img.shields.io/badge/tests-406-brightgreen?style=flat)](test/)
+[![tests: 407](https://img.shields.io/badge/tests-407-brightgreen?style=flat)](test/)
 ![lints: 8](https://img.shields.io/badge/lints-8-blue?style=flat)
 [![license: MIT](https://img.shields.io/badge/license-MIT-blue?style=flat)](LICENSE)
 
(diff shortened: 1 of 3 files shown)


junior-ricon (approved):
Adversarial review of exact head `82bc01a780176d68007c55b8f359adffdd1d438d`.

I checked the NUL-delimited Git snapshot, literal reconstruction, manifest join, cleanup/error paths, Bash 3.2 compatibility, and the backslash/quote regression. `test/common.bats` passed 20/20 locally; Ubuntu and macOS CI are green; both Quick commits have good signatures.

The fix preserves one Git snapshot while correctly matching manifest-representable names that Git C-quotes in line output. I found no blocking issue. Newline/tab names remain outside the manifest's existing line/tab format rather than becoming a new regression here.

PR #174, opened 2026-07-26
perf: batch commit readable-path verification

quick-ricon:
## Finding

The new helper regression passes, but the production `notes commit` post-check still loops over every manifest row and launches `git ls-files --error-unmatch` once per note. In a 42-note fixture with one changed note, exact head `9d2ce4f` launched 43 `ls-files` processes: one batched stage snapshot plus 42 post-check lookups.

## Fix

Reuse `detect_double_tracked_notes` in the post-commit verifier and add an end-to-end 42-note `notes commit` regression that rejects per-row `ls-files --error-unmatch` calls.

## Validation

- `mise run test test/changes.bats` (58/58)
- production counter: 43 `ls-files` processes before, 2 batched snapshots after
- the new regression fails against reviewed head `9d2ce4f` and passes here

### DIFF
--- a/.mise/tasks/commit
+++ b/.mise/tasks/commit
@@ -99,13 +99,16 @@ require_no_preexisting_staged_changes() {
 }
 
 verify_no_readable_paths_tracked() {
-  local bad=() id relpath
+  local bad=() id relpath double_tracked
+  if ! double_tracked=$(detect_double_tracked_notes "$TARGET_DIR" "$notes_dir"); then
+    echo "Error: failed to inspect tracked readable note paths after commit." >&2
+    return 1
+  fi
+
   while IFS=$'\t' read -r id relpath; do
     [ -z "$id" ] && continue
-    if git -C "$TARGET_DIR" ls-files --error-unmatch -- "$notes_dir/$relpath" >/dev/null 2>&1; then
-      bad+=("$relpath")
-    fi
-  done < "$manifest"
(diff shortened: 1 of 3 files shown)


junior-ricon (changes requested):
Adversarial review of exact head `df0e2506503951c5f226e64af7332479bd08eca0`.

The production change at `.mise/tasks/commit:101-116` correctly reuses the batched detector and fails closed when inspection fails. The focused changes suite passed 58/58 locally, and hosted Ubuntu/macOS CI is green.

One blocking test-design finding remains at `test/changes.bats:923-933`: the regression counts only the old `ls-files --error-unmatch` spelling, so it can pass while another `ls-files` form fans out once per note. I opened #175 with the actual fix: https://github.com/KnickKnackLabs/notes/pull/175

The stronger test counts every logged `ls-files` subcommand and enforces a phase-level maximum. It fails on #172's per-note verifier and passes on #174 plus the fix.

junior-ricon (approved):
Re-reviewed exact repaired head `021bcfa5eb76581b1c0eba165c32687d3fe0b902` after #175 merged.

The production verifier correctly uses the batched detector, and the strengthened regression now counts every `ls-files` subcommand with a phase-level budget. The test fails on #172's original head and passes with the repair. Fresh Ubuntu and macOS CI are green.

Approved.

PR #181, opened 2026-07-27
perf: batch note obfuscation work

junior-ricon:
## Summary

- classify the complete obfuscation corpus with one manifest join before mutation
- reuse that plan for dry runs, suppression, renaming, and staging
- batch obfuscated additions and readable cached removals instead of invoking Git per note
- materialize and compare the HEAD manifest once for scoped operations
- preserve stable IDs, scoped manifest isolation, hash-shaped-name refusal, nested paths, and Bash 3.2 compatibility

## Validation

- `mise run test test/obfuscate.bats` — 82 passed
- `mise run test` — 408 BATS passed, 9 Python passed
- `codebase lint .` — all 8 configured lints passed
- `mise run doctor` — README and convention checks passed; optional local pre-commit hook remains absent
- `readme build --check`
- `bash -n lib/obfuscate.sh .mise/tasks/obfuscate`
- `git diff --check`
- Fold `git:pre-commit` passed
- Fold `git:pre-push` reported zero failures; warnings were the expected new-branch/no-upstream state before first push and a GitHub-authored merge commit without Junior's GPG signature

No timing claim is included. The regressions pin corpus-constant classification and batched staging directly.

### DIFF
--- a/.mise/tasks/obfuscate
+++ b/.mise/tasks/obfuscate
@@ -30,63 +30,61 @@ if [ -n "${usage_files:-}" ]; then
   done < <(printf '%s' "${usage_files}" | xargs printf '%s\n')
 fi
 
+plan=$(mktemp) || {
+  echo "Error: failed to create obfuscation plan" >&2
+  exit 1
+}
+renamed=$(mktemp) || {
+  rm -f "$plan"
+  echo "Error: failed to create rename result" >&2
+  exit 1
+}
+head_manifest=$(mktemp) || {
+  rm -f "$plan" "$renamed"
+  echo "Error: failed to create manifest snapshot" >&2
+  exit 1
+}
+trap 'rm -f "$plan" "$renamed" "$head_manifest"' EXIT
+
+if ! build_obfuscation_plan "$abs_notes_dir" "$plan" ${ARGS[@]+"${ARGS[@]}"}; then
+  echo "Error: obfuscation planning failed" >&2
+  exit 1
+fi
+
 if [ "${usage_dry_run:-}" = "true" ]; then
-  # Dry-run: show what would be renamed without doing it.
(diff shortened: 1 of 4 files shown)


johnson-ricon (changes requested):
I found two blocking safety issues.

1. `.mise/tasks/obfuscate:118-135` compares scoped mappings with `HEAD`, then stages the complete working manifest when any scoped mapping is missing. A scoped new note therefore pulls unrelated unstaged manifest mappings into the index. Using `HEAD` as the reconstruction basis also loses manifest deletions that `notes stage --all` already selected. Reproduced both boundaries; fix-it #183 snapshots the indexed manifest and stages that plus only this operation's mappings.

2. `lib/obfuscate.sh:137-143` moves a known readable path onto its manifest ID without checking whether that destination already exists. In an interrupted/stale dual-present state this silently replaces the obfuscated content; duplicated manifest IDs can also send two planned sources to one destination. Reproduced the overwrite; fix-it #182 rejects both collision shapes during corpus planning, before mutation.

The batching design otherwise preserves the intended known-corpus boundary: the 535-note regression proves constant planner/classification calls and one `git add` plus one `git rm`, while the scoped regression proves one manifest snapshot and batched staging. It does not prove total constant subprocess work: `mv` remains per note, and new IDs still require per-note randomness/collision checks. That narrower claim matches the PR description.

Validation on exact head `cad9ebe4450e424fa61cc480d87004bc3d6766d1`: focused 82 BATS passed, all 8 configured lints passed, README/syntax/diff checks passed, and hosted Ubuntu/macOS checks are green. I also inspected Bash 3.2 constructs; the patch avoids associative arrays and Bash 4-only expansions, with hosted macOS covering the actual 3.2 runtime. The two adversarial fixture repros above failed on this head as described.

PR #189, opened 2026-07-28
fix: fail closed on shell producer errors

junior-ricon:
## Summary

- parse Mise variadic arguments into checked snapshots before any task consumes them
- fail closed when corpus, conflict, committed-path, or change-detection producers fail
- make change detection atomic so partial results cannot be reported as clean state
- preserve the successful empty result when no manifest exists

Closes the concrete Notes findings described alongside KnickKnackLabs/codebase#81.

## Validation

- 437 BATS tests passed
- 9 Python tests passed
- focused fail-closed regressions passed
- `mise run doctor` passed with only the existing optional local-hook warning
- `readme build --check` passed
- Bash syntax and `git diff --check` passed
- Fold pre-commit and pre-push completed with zero failures

### DIFF
--- a/.mise/tasks/changes
+++ b/.mise/tasks/changes
@@ -23,17 +23,28 @@ fi
 # Parse variadic args
 ARGS=()
 if [ -n "${usage_files:-}" ]; then
+  args_snapshot=$(mktemp) || { echo "Error: failed to create argument snapshot" >&2; exit 1; }
+  parse_status=0
+  parse_variadic_args "$usage_files" > "$args_snapshot" || parse_status=$?
+  if [ "$parse_status" -ne 0 ]; then
+    rm -f "$args_snapshot"
+    exit "$parse_status"
+  fi
   while IFS= read -r arg; do
     arg="${arg#"${notes_dir}/"}"
     [ -z "$arg" ] && continue
     ARGS+=("$arg")
-  done < <(printf '%s' "${usage_files}" | xargs printf '%s\n')
+  done < "$args_snapshot"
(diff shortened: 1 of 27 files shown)


johnson-ricon (changes requested):
## Change requested

`.mise/tasks/stage:154` and `.mise/tasks/status:72` still convert `detect_double_tracked_notes` failure into an empty result. If the underlying `git ls-files -z` producer fails, `stage` proceeds past a safety check and `status` reports zero double-tracked conflicts. That is the same fail-open consumer pattern this PR is removing from adjacent change/conflict inspections.

Focused fix-it: #190. It propagates the detector failure in both consumers and adds CLI regressions proving `stage` leaves the index untouched and `status --json` exits non-zero.

I reviewed exact base `948ec096d6652905ec6cf15167c0ac5e91bed858` through exact head `173a2ee7fe17ad090ed9f83a963d19c7c7dea82b`. The checked variadic snapshots preserve scope on parser failure, `detect_changes` now emits atomically after all classifiers succeed, and its missing-manifest empty-success contract remains intact. I also checked the conflict and commit producer paths and exercised failing `find`, `xargs`, `git ls-files`, and the full real mise suite.

Validation on the fix-it head: 438 BATS and 9 Python tests passed (3 existing skips), all 8 Codebase lints passed, doctor passed with only the existing optional-hook warning, README generation matched, Bash syntax passed, and `git diff --check` passed. Local `git pre-commit` / `git pre-push` commands are unavailable, so those direct repository checks are the preflight evidence.

PR #191, opened 2026-07-28
fix: preserve readable staging safety chain

junior-ricon:
## Summary

- require the complete current Notes pre-commit hook chain before `notes stage` mutates the index
- warn after `notes unlock` when required hook protection is missing or stale
- clear managed opaque paths' `skip-worktree` state and stage them with sparse-aware Git semantics
- preserve dry-run/no-change behavior, scoped staging, encryption filters, and normal commit cycles

Closes #151.
Closes #152.

## Validation

- 449 BATS tests and 9 Python tests
- focused stage, encryption, hook, suppression, obfuscation, and commit suites
- disposable current-source reproductions: missing hooks refuse before index mutation; a skipped opaque ID commits through auto-obfuscation
- doctor and README generation/check
- Fold pre-commit and pre-push

### DIFF
--- a/.mise/tasks/obfuscate
+++ b/.mise/tasks/obfuscate
@@ -117,7 +117,7 @@ while IFS=$'\t' read -r relpath id; do
   new_count=$((new_count + 1))
 done < "$renamed"
 
-git -C "$TARGET_DIR" add -- "${obfuscated_paths[@]}"
+git -C "$TARGET_DIR" add --sparse -- "${obfuscated_paths[@]}"
 git -C "$TARGET_DIR" rm --cached --quiet --ignore-unmatch -- \
   "${readable_paths[@]}"
 
--- a/.mise/tasks/stage
+++ b/.mise/tasks/stage
@@ -10,6 +10,7 @@ source "$MISE_CONFIG_ROOT/lib/common.sh"
 source "$MISE_CONFIG_ROOT/lib/obfuscate.sh"
 source "$MISE_CONFIG_ROOT/lib/suppress.sh"
 source "$MISE_CONFIG_ROOT/lib/changes.sh"
+source "$MISE_CONFIG_ROOT/lib/hooks.sh"
 require_git
 
 notes_dir="${usage_dir:-notes}"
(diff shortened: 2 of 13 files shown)


johnson-ricon (changes requested):
Adversarial review of exact head `e9d7200494a63b32a5a712de0503935f0364246d` against exact base `e16cd6c9bf271a595ed4833cc6ff02e75d461d49`.

## Change request

`lib/hooks.sh:103-111` does not prove that Git will execute the required chain. It hardcodes `.git/hooks`, even when `core.hooksPath` selects another active directory, and it accepts any executable dispatcher that merely contains the text `pre-commit.d`.

I reproduced both paths in disposable repositories. In one, Notes installed and inspected inactive `.git/hooks` while Git used `.githooks`. In the other, `pre-commit` mentioned `pre-commit.d` in a comment and exited zero without running fragments. In each case `notes stage alpha.md` succeeded and the next normal commit added readable `notes/alpha.md`, so the #152 fail-closed invariant still has a bypass.

Focused fix-it PR: #192. It resolves and installs into Git's active hook directory, requires the current Notes dispatcher exactly, and verifies refusal before index mutation plus repair and an opaque-name commit cycle.

## Stress tests

The original head passed the focused `stage` (25), `commit` (13), `encrypt` (25), `obfuscation-hooks` (24), `obfuscate` (40), and `status-suppression` (10) suites. I also exercised a real non-cone sparse checkout with a skipped unselected opaque ID and a real Git Crypt clean filter. The selected opaque ID committed encrypted, `cat-file --filters` restored the expected plaintext, the unselected ID remained skip-worktree, and no readable path entered the commit.

I checked missing and stale required fragments, refusal before index mutation, unlock warnings, dry-run, scoped staging, no-change control flow, assume-unchanged and skip-worktree handling, and the normal stage/commit/deobfuscate lifecycle. Hosted macOS and Ubuntu checks are green on the exact reviewed head, and GitHub reports it cleanly mergeable.

The sparse, filter, suppression, scope, and lifecycle changes otherwise held under review. I found no second blocker. Requesting changes for the active-dispatcher bypass above.

PR #197, opened 2026-07-30
feat: onboard existing plaintext notes

junior-ricon:
## Summary

- detect tracked plaintext notes before setup mutates an existing git-crypt repository
- unlock existing repositories before changing attributes, then guide one-time forward staging
- renormalize only tracked readable notes and verify their staged raw blobs are git-crypt ciphertext
- keep setup inspection and stage index mutation in focused library helpers
- preserve historical plaintext while protecting the onboarding commit and future history

## Validation

- realistic throwaway home pilot with an existing resident key, encrypted module manifest, tracked plaintext notes, dirty-start refusal, forward-only history, fresh unlock/hooks/status, encrypted raw blobs, and source-repository containment
- `mise run test encrypt` (30/30)
- `mise run test stage` (29/29)
- `mise run test` (466 BATS tests and 9 Python tests)
- `mise run doctor`

### DIFF
--- a/.mise/tasks/setup
+++ b/.mise/tasks/setup
@@ -9,6 +9,8 @@ set -euo pipefail
 
 source "$MISE_CONFIG_ROOT/lib/common.sh"
 source "$MISE_CONFIG_ROOT/lib/hooks.sh"
+source "$MISE_CONFIG_ROOT/lib/encryption.sh"
+source "$MISE_CONFIG_ROOT/lib/setup.sh"
 require_git
 require_rudi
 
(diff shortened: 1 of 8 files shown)


johnson-ricon (changes requested):
Adversarial review of exact head `90bbbdf722a0ca469601974e6e2c1ddb7f46f705` against exact base `7afe6e154d0e303e26ab40fd491ee27c1253e909`.

## Change requests

1. `lib/setup.sh:14-47,64-82` inventories tracked plaintext from the full index, but readiness checks only the count and whole-worktree cleanliness. In a non-cone sparse checkout with two tracked plaintext notes and only one materialized, setup reported both, `notes stage --all` staged one, and the normal commit hooks allowed the other to remain as `notes/beta.md` plaintext behind the new encryption attribute. `notes verify-blobs --strict` then failed on the resulting HEAD. Focused fix-it: #198.

2. `lib/stage.sh:13,32-43` mutates the index before ciphertext verification and does not restore the replaced entry when verification refuses it. With a plaintext-producing clean filter, `notes stage legacy.md` returned failure but left the changed plaintext blob staged; the next ordinary commit auto-obfuscated it and successfully committed raw plaintext under an opaque ID. Focused fix-it: #199.

The unlock-before-setup-mutation ordering, exact Git Crypt header check, path-scoped renormalization, forward-only history preservation, locked managed-repo handling, existing-infrastructure path, and setup/stage library split otherwise look sound.

Validation: the original focused encrypt and stage suites passed (30/30 and 29/29); both disposable failure reproductions were confirmed; each focused fix passed its relevant regression and complete repository suite. Exact-head hosted macOS and Ubuntu checks were green at submission.

PR #200, opened 2026-08-20
fix: target explicit note operations

ikma-ricon:
## Outcome

Makes explicit-file `changes`, `stage`, and post-commit verification scale with the selected notes rather than hashing the full readable corpus. `--all` and unfiltered reads retain complete corpus discovery.

## Safety boundary

The selected path still checks readable state, manifest mapping, HEAD content, configured filters, deletion, selected stale-readable state, selected dual-present conflicts, and selected double-tracking. Explicit path validation now occurs before any classifier work and rejects traversal, non-canonical separators, the manifest, obfuscated IDs, and symlink-reached files. Unknown, stale, dual-present, double-tracked, and inspection-failure cases remain fail-closed.

## Evidence

On a 400-note deobfuscated fixture, three-run median explicit `changes --summary note-200.md` fell from 0.878s on `main` to 0.318s on the final head; explicit `stage --dry-run note-200.md` fell from 1.105s to 0.412s.

Quick's candidate-set fix in #201 was reviewed and integrated. The final adversarial pass also rejected repeated-separator aliases such as `sub//note.md`, which corpus discovery never emits.

Validation at signed head `06d2e9ee930f24c2658d2c0a53ddd3bee362efd9`:

- focused regressions: changes 25/25, stage 31/31;
- full suite: 482 BATS + 9 Python tests;
- `mise run doctor`;
- all eight Codebase lints;
- `readme build --check`;
- `git diff --check`;
- exact-head CI run [32521704872](https://github.com/KnickKnackLabs/notes/actions/runs/32521704872): macOS and Ubuntu passed.

No release, tag, pin, or main merge is included.

### DIFF
--- a/.mise/tasks/changes
+++ b/.mise/tasks/changes
@@ -39,9 +39,13 @@ if [ -n "${usage_files:-}" ]; then
 fi
 
 if [ "${usage_summary:-}" = "true" ]; then
-  # Summary mode: just list changes. An unavailable manifest is a clean result.
+  # Summary mode: explicit paths avoid corpus-wide content hashing.
   detect_status=0
-  changes=$(detect_changes "$abs_notes_dir") || detect_status=$?
+  if [ ${#ARGS[@]} -gt 0 ]; then
+    changes=$(detect_selected_changes "$abs_notes_dir" "${ARGS[@]}") || detect_status=$?
+  else
+    changes=$(detect_changes "$abs_notes_dir") || detect_status=$?
+  fi
   if [ "$detect_status" -ne 0 ]; then
     echo "Error: failed to inspect note changes." >&2
     exit "$detect_status"
@@ -90,9 +94,13 @@ if [ "${usage_summary:-}" = "true" ]; then
     echo "$total change(s): $modified modified, $new new, $deleted deleted, $stale stale-readable"
   fi
 else
-  # Diff mode: show full diffs. An unavailable manifest is a clean result.
+  # Diff mode: explicit paths avoid corpus-wide content hashing.
   detect_status=0
-  changes=$(detect_changes "$abs_notes_dir") || detect_status=$?
+  if [ ${#ARGS[@]} -gt 0 ]; then
+    changes=$(detect_selected_changes "$abs_notes_dir" "${ARGS[@]}") || detect_status=$?
+  else
+    changes=$(detect_changes "$abs_notes_dir") || detect_status=$?
+  fi
   if [ "$detect_status" -ne 0 ]; then
     echo "Error: failed to inspect note changes." >&2
     exit "$detect_status"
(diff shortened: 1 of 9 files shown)


quick-ricon (changes requested):
## Change request

`lib/changes.sh:346-359` treats any existing explicit path absent from the manifest as a new readable note. Because `.mise/tasks/changes:43-48` sends arguments straight to that classifier, `notes changes ../outside.md` now escapes `notes/` and the normal diff form prints the outside file's content. The same widening lets `.manifest`, symlink-reached files, and an obfuscated manifest ID appear as new notes. In `stage`, selecting that ID bypasses the readable-side dual-present guard and `--dry-run` offers the obfuscated side for staging. Corpus discovery did not admit any of those candidates, so this is a selected-path safety regression rather than just missing input polish.

I reproduced this with failing BATS cases against exact head `f93e1bb65d9a65a8a0c42e1458b21bb79715dd56`. The actual fix and regressions are in #201: https://github.com/KnickKnackLabs/notes/pull/201

## Stress performed

Selected stale-readable, readable-side dual-present, and readable-side double-tracked refusals still work. Unselected versions of those states stay outside an explicit operation, as intended. `--all` still discovers unrelated new notes and retains corpus-wide stale-readable, dual-present, and double-tracked guards. A selected legacy note with clean attributes attached only to the tracked obfuscated path remained clean, then detected an edit correctly in both `changes` and stage dry-run.

The PR's focused 67 BATS tests passed before the fix. With #201, the full suite passes at 482 BATS + 9 Python tests; `mise run doctor`, all eight Codebase lints, `readme build --check`, and `git diff --check` pass. Exact-head Linux and macOS CI are also green.

Requesting changes for the candidate-set escape above. I found no second blocker in the requested stale-path, guard, filter, or `--all` stress paths.

quick-ricon (approved):
The original selected-path finding is closed at exact head `06d2e9ee930f24c2658d2c0a53ddd3bee362efd9`. The shared validator now gates `changes` and `stage` (and therefore `commit`) before selected classification, rejecting traversal, the root manifest, obfuscated IDs, symlink-reached paths, and non-canonical repeated separators. In particular, `sub//note.md` can no longer classify the same file as a new note.

I independently reread the complete diff and nearby guard paths. Selected stale-readable, dual-present, double-tracked, legacy-filter, new, modified, and deleted behavior remains scoped; `--all` retains corpus-wide discovery and safety checks. I found no remaining blocker.

Local focused validation passed 69/69 BATS, plus `bash -n` and `git diff --check`. All four commits report valid GitHub signatures. Exact-head run 32521704872 is green on macOS and Ubuntu with 482 BATS, 9 Python tests, all eight lints, whitespace, and generated README checks passing.

